PR #84, opened 2026-05-22
feat(eval): pipeline diagnostic & retry hardening

yilu331:
Plan A of the /evaluations redesign work. Makes the existing evaluation pipeline observable and self-healing — AND fixes a silent product bug discovered during live e2e verification.

## Summary

- **EvalHealth singleton** tracks scheduler ticks, per-reason skip counts, and rolling 24h producer-failure counts in one process-wide store.
- **Producer retries** `save_agent_success_evaluation_results` with backoff (1s / 4s, 3 total attempts); records final failures into EvalHealth instead of silently dropping them.
- **Scheduler loop** records its tick into EvalHealth on every iteration.
- **All five skip paths** in `run_group_evaluation` bump matching `SkipReason` counters so the operator can see why sessions are dropping.
- **New `/healthz/eval` endpoint** surfaces all three signals with a green/amber/red liveness derivation (>5min silent = amber, >30min = red).
- **Bug fix — agentic-path scheduler bypass.** The agentic extraction backend hit `return result` before reaching the scheduler invocation in the classic path, so no AgentSuccessEvaluationResult records were ever produced for any org running with `extraction_backend=agentic`. This was the silent root cause of empty /evaluations tiles. Extracted into a shared `_schedule_group_evaluation_if_needed` helper called from both code paths.

## Live verification

Discovered and fixed during local e2e testing — published 10 substantive sessions to a backend running this branch, observed:
- Before the agentic fix: `last_tick_monotonic: null`, `liveness: red`, no scheduler logs.
- After the fix: `Scheduled group evaluation for key=...` → `Firing group evaluation` → `Running group evaluation for session=...` in logs; `/healthz/eval` reports `liveness: green` with populated tick.

## Commits

1. `EvalHealth singleton for pipeline diagnostics`
2. `scheduler records tick into EvalHealth`
3. `bump SkipReason counters in each skip path`
4. `retry save with backoff and record producer failures`
5. `/healthz/eval exposes pipeline diagnostics`
6. `schedule group evaluation from the agentic path too` *(discovered via e2e)*

## Test plan

- [x] Unit tests for EvalHealth counters, 24h decay window, tick recording
- [x] Skip-reason counter tests for `run_group_evaluation`
- [x] Save-retry tests: 3 attempts with backoff = [1s, 4s], success on attempt 2
- [x] `/healthz/eval` route tests for green/amber/red liveness derivation
- [x] Regression test for `_schedule_group_evaluation_if_needed`
- [x] Full submodule regression: 127 tests pass
- [x] Live e2e against running backend: scheduler fires through to runner; /healthz/eval reports green

## Companion PR

ReflexioAI/reflexio-enterprise#100 bumps the submodule pointer here.

## Operational follow-up (out of scope for Plan A)

During live e2e, the runner logged `No agent_success_evaluation extractor configs found` — the deployed config has `agent_success_configs: null`, so even when the scheduler fires correctly, the evaluator produces no records. This is a config gap, not a code bug, and is separate from Plan A.

### DIFF
--- a/reflexio/server/api_endpoints/health_api.py
+++ b/reflexio/server/api_endpoints/health_api.py
@@ -15,6 +15,8 @@ from typing import Any
 
 from fastapi import FastAPI, Request, Response
 
+from reflexio.server.services.agent_success_evaluation import _eval_health
+
 _STARTED_AT = time.monotonic()
 _REQUEST_COUNT = 0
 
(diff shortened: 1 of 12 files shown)


yilu331:
Superseded by #91 — consolidated evaluations + Braintrust PR. Branch remains on remote for history.

PR #90, opened 2026-05-23
feat(braintrust): Plan C-cron — recurring 15-min sync scheduler

yilu331:
Unblocks unattended Braintrust sync. Singleton `BraintrustSyncScheduler` daemon thread walks every connected org every 15 min (5s under `IS_TEST_ENV`). Also adds `BRAINTRUST_BASE_URL` env override flowing through both cron + manual /sync paths via a unified client factory. 14 new tests pass.

## Depends on
`reflexio#86` + `reflexio#87` + `reflexio#88` + `reflexio#89` (chain of B-backend, C-backend, C-overview, SQLite storage).

## Out of scope
- Lifecycle wiring (calling `scheduler.start()` from the app's startup hook). Currently the scheduler is constructible but not auto-started; the embedding application controls that. Plan C-frontend or a small startup PR will wire it.

### DIFF
--- a/reflexio/lib/_dashboard.py
+++ b/reflexio/lib/_dashboard.py
@@ -1,4 +1,13 @@
 from reflexio.lib._base import STORAGE_NOT_CONFIGURED_MSG, ReflexioBase
+from reflexio.models.api_schema.eval_overview_schema import (
+    ContextTile,
+    GetEvaluationOverviewRequest,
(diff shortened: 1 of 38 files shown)


yilu331:
Superseded by #91 — consolidated evaluations + Braintrust PR. Branch remains on remote for history.

PR #126, opened 2026-06-05
feat(billing): OSS metering foundation (usage signals + DI seams)

yilu331:
## Summary

Adds the **OSS metering foundation** that the (closed-source) billing layer builds on. This is the open-source half: it emits structured usage signals and exposes clean dependency-injection seams so a downstream deployment can attach billing/enforcement **without the OSS core depending on any of it**. With no overrides registered, behavior is unchanged — every hook is a no-op and metering is fully isolated from the request success path.

Three value signals are emitted as `UsageEvent`s:

1. **Learning** — extraction input tokens (input-anchored, model-agnostic `cl100k_base` count) + learnings generated. Metered **only** for profile/playbook generation; eval/reflection pipelines do not bill.
2. **Application** — applied learnings (surfaced top-K) for production agents only; dashboard/non-agent callers and empty results emit nothing.
3. **Access** — caller-type classification seam (`default_get_caller_type`, no-op `"internal"` in OSS).

## Changes

- **`billing_signals.py`** — single source of the platform-LLM rule + the canonical billing tokenizer (`count_input_tokens`).
- **`billing_meter.py`** — emission helpers; the one source of usage-event shape.
- **`usage_metrics.py`** — `UsageEvent` gains token/attribution/caller fields; `record_usage_event` swallows recorder failures so metering can never 500 a request.
- **`llm/token_accounting.py`** — `RunTokenTotals` + `sum_trace_tokens` fold tool-loop trace tokens into run totals reachable by the generation service.
- **`services/base_generation_service.py`** — `EMITS_LEARNING_BILLING` flag (true only on profile/playbook generation); emission isolated in try/except.
- **`server/_auth.py` / `server/api.py`** — `default_get_caller_type` + no-op `default_billing_gate` DI seams; `create_app(get_caller_type=, get_billing_gate=)`; applied-learnings metering wired into the 5 search/read endpoints; `get_agent_playbooks` gains a `120/minute` rate limit.

## Isolation guarantees

- No new dependency on any closed-source / enterprise code.
- Every metering call site is wrapped so a failure logs and continues — the product response is never affected (covered by `test_metering_failure_does_not_break_search_response`).
- All hooks default to no-ops; a deployment opts in via `app.dependency_overrides`.

## Test plan

- `tests/server/test_billing_signals.py`, `tests/server/test_billing_meter.py` — emission helpers + tokenizer (empty-string, encoding-drift guard, special-token handling).
- `tests/server/llm/test_token_accounting.py` — None-coercion, empty/missing-trace guards, multi-turn summation.
- `tests/server/api_endpoints/test_applied_learnings_metering.py` — per-endpoint surfaced-count metering across all 5 endpoints × {production-agent, dashboard, empty}.
- `tests/server/services/test_generation_billing_emission.py` — learning emission gated to profile/playbook generation; eval pipeline does not bill.
- 36 metering/billing tests pass; `ruff` + `pyright` clean; `import reflexio` verified.

## Summary by CodeRabbit

* **New Features**
  * Metering for learnings (generated and applied) with per-run token accounting, optional generation billing, and configurable per-line billing gates with overridable caller-type classification.
  * Usage events now include prompt/completion/input token counts, platform/caller context, and accept request/session correlation IDs for billing attribution; app factory can toggle mounting of data-plane routes.

* **Tests**
  * Added comprehensive tests for applied‑learnings metering, token accounting, billing signal detection, generation billing emission, and app router mounting behavior.

### DIFF
--- a/reflexio/models/api_schema/retriever_schema.py
+++ b/reflexio/models/api_schema/retriever_schema.py
@@ -50,6 +50,11 @@ class SearchInteractionRequest(BaseModel):
 class SearchUserProfileRequest(BaseModel):
     user_id: NonEmptyStr
     generated_from_request_id: str | None = None
+    # Caller correlation IDs for billing attribution on the Application line.
+    # Optional; when populated they flow into the usage event request_id /
+    # session_id columns via _meter_applied_learnings in server/api.py.
+    request_id: str | None = None
+    session_id: str | None = None
     query: str | None = None
     start_time: datetime | None = None
     end_time: datetime | None = None
@@ -233,6 +238,10 @@ class GetAgentPlaybooksRequest(BaseModel):
     agent_version: str | None = None
     status_filter: list[Status | None] | None = None
     playbook_status_filter: PlaybookStatus | None = None
+    # Caller correlation IDs for billing attribution on the Application line.
+    # Optional; consumed by _meter_applied_learnings in server/api.py.
+    request_id: str | None = None
+    session_id: str | None = None
 
 
 class GetAgentPlaybooksResponse(BaseModel):
@@ -267,6 +276,10 @@ class SearchUserPlaybookRequest(BaseModel):
     threshold: float | None = Field(default=0.4, ge=0.0, le=1.0)
     enable_reformulation: bool | None = False
     search_mode: SearchMode = SearchMode.HYBRID
+    # Caller correlation IDs for billing attribution on the Application line.
+    # Optional; consumed by _meter_applied_learnings in server/api.py.
+    request_id: str | None = None
+    session_id: str | None = None
 
     @model_validator(mode="after")
     def check_time_range(self) -> Self:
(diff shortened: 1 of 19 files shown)


yilu331:
Thanks @coderabbitai — all three nitpicks addressed in `9044c73`:

- **`tests/server/api_endpoints/test_applied_learnings_metering.py` — add a path for the `get_billing_gate` override seam** (⚡ Quick win): done. Extended `_client` to accept `get_billing_gate`, and added `test_get_billing_gate_override_seam`, which asserts `create_app` wires an override for **both** billing lines (`application` + `learnings_generated`) keyed by the lru-cached `default_billing_gate` sentinel, and that the `application` override actually **fires** on a real `/api/search` request. (The `learnings_generated` line is pinned via the override map rather than by driving the heavier `/api/publish_interaction` handler — same DI-contract protection, less brittle.)
- **`reflexio/server/api.py` — misleading `_line` underscore** (💤 Low value): renamed `_line` → `line`; the variable is used in the loop body, so the underscore prefix was wrong.
- **`reflexio/server/llm/token_accounting.py:28-29` — redundant `int()` wrapper** (💤 Low value): keeping this one, intentionally. `RunTokenTotals.add()` is fed from `sum_trace_tokens`, which reads `prompt_tokens` / `completion_tokens` off an **untyped** trace object via `getattr(turn, ..., None)` — so a non-`int` (e.g. a float token count from a provider trace) can reach `add()` despite the `int | None` annotation. `int(x or 0)` keeps the running totals integral. It's a deliberate defensive guard, not redundant code.

ruff + pyright clean on both files; 17/17 metering tests pass.

PR #181, opened 2026-06-19
Lineage Phase A — content-free provenance + forward-resolution (OSS/SQLite)

yilu331:
## Lineage Phase A — OSS / SQLite half

Adds a **content-free lineage/provenance layer** so that when a user playbook / agent playbook / profile is **merged or superseded**, references resolve **forward to the survivor** instead of dangling. This is the OSS (shared, SQLite) half; the enterprise Postgres+Supabase half is in the corresponding `reflexio-enterprise` PR.

### What's here
- **Data model:** `MERGED`/`SUPERSEDED` statuses; `merged_into`/`superseded_by` pointers on the three entity types; a content-free, append-only `lineage_event` table (`UNIQUE(entity_id, op, request_id)` idempotency); `LineageEvent`/`LineageContext`/`RecordRef` models.
- **Atomic primitive:** `merge_records` / `supersede_record` — row mutation + forward pointer + one `lineage_event`, in a single `with self._lock:` transaction (no `_execute_atomic` helper exists; uses the `delete_user_playbooks_by_ids` pattern).
- **Write paths routed through it:** the consolidator `unify` merge, the shared `apply_playbook_edit` (reflection + tuner), and the optimizer supersede — each now atomic, lineage-recording, and orphan-free. `hard_delete` audit events emitted before physical deletes.
- **Reads:** universal base status filter hides tombstones from default reads (NULL-safe — CURRENT rows always shown), with `include_tombstones=True` on the by-id getters; `resolve_current()` follows the pointer to the live survivor (1-hop, cycle-safe, purged-aware).

### Incidental fixes found during the build
- `apply_playbook_edit` no longer leaves an orphan CURRENT row on a lost race.
- The `ARCHIVED → SUPERSEDED` shift fixes a latent `run_downgrade` bug that would restore reflection-archived rows to CURRENT.

### Testing
TDD throughout — 11 tasks, each with integration tests (real SQLite storage). New lineage suites + a parametrized storage contract test all green; broader storage/reflection/optimizer/consolidator suites pass with no regressions.

### Notes
- Foundation commits (`apply_playbook_edit` extraction, `PlaybookRetrievalLog`/`RetrievalLogMixin`) are included as the base this builds on.
- Per-spec deferred to Phase B/C: in-place `update_*` instrumentation, TTL GC, legacy-log retirement, multi-hop path-compression, the agent-aggregate erasure-linkage question.

Spec/plan live in the enterprise repo (`docs/superpowers/specs|plans/…lineage…`).

## Release Notes

* **New Features**
  * Playbook consolidation now records merge/revise lineage end-to-end, including persisted lineage events and retrieval-log support.
  * Lineage-aware record replacement uses atomic supersede to maintain consistent “current” state.
  * Lineage resolution can follow merge chains to determine the current live record.

* **Improvements / Behavior Changes**
  * Storage reads are tombstone-aware by default (merged/superseded records are hidden unless explicitly included).
  * Hard-deleting playbooks can emit erasure lineage events.
  * Lineage-based anomaly reporting was improved for cycle/hop-resolution edge cases.

### DIFF
--- a/reflexio/models/api_schema/domain/entities.py
+++ b/reflexio/models/api_schema/domain/entities.py
@@ -119,6 +119,9 @@ __all__ = [
     "AdminInvalidateCacheRequest",
     "AdminInvalidateCacheResponse",
     "PlaybookRetrievalLog",
+    "LineageEvent",
+    "LineageContext",
+    "RecordRef",
 ]
 
 # ===============================
@@ -227,6 +230,8 @@ class UserProfile(BaseModel):
     source_span: str | None = None
     notes: str | None = None
     reader_angle: str | None = None
+    merged_into: str | None = None
+    superseded_by: str | None = None
 
 
 # user playbook for agents
@@ -252,6 +257,8 @@ class UserPlaybook(BaseModel):
     source_span: str | None = None
     notes: str | None = None
     reader_angle: str | None = None
+    merged_into: int | None = None
+    superseded_by: int | None = None
 
 
 class ProfileChangeLog(BaseModel):
@@ -281,6 +288,8 @@ class AgentPlaybook(BaseModel):
     status: Status | None = (
         None  # used for tracking intermediate states during playbook aggregation. Status.ARCHIVED for playbooks during aggregation process, None for current playbooks
     )
+    merged_into: int | None = None
+    superseded_by: int | None = None
 
 
 class PlaybookOptimizationJob(BaseModel):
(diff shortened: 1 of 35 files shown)


yilu331:
Thanks for the review — addressed in `69fa5a7`, and the branch has been rebased onto latest `main` (so the diff is now conflict-free).

**Fixed in this push (`69fa5a7`):**
- **Idempotency key now includes `entity_type`** → `UNIQUE (org_id, entity_type, entity_id, op, request_id)` (both SQLite DDL spots, the `append_lineage_event` fallback `SELECT`, and the abstract-base contract; the enterprise PL/pgSQL migration is mirrored in reflexio-enterprise#284). This closes the cross-entity-type collision flagged in `_base.py` and `storage_base/_lineage.py` — two entity types sharing a numeric id no longer dedupe each other within an org.
- **`merge_records` survivor guard** — a `source_id` equal to `survivor_id` is now skipped (and excluded at the SQL `WHERE`), so the survivor can never be tombstoned into itself.
- **`get_lineage_events` now reads under `self._lock`** (it was bypassing the lock on the shared `check_same_thread=False` connection).
- **`resolve_current(... id)` → `record_id`** (no builtin shadow).
- **Test lint:** split the semicolon-chained statements (E702) and dropped the unused unpack (RUF059).

**Already addressed by earlier commits now on the branch** (CodeRabbit's re-review should reconcile these):
- *Rollback cleanup audited as a logical hard delete* (`_playbook.py`) — `delete_user_playbooks_by_ids` / `delete_agent_playbooks_by_ids` now take `emit_hard_delete=False`, passed by the supersede / `apply_playbook_edit` rollback paths, so a never-CURRENT successor cleanup no longer emits a phantom `hard_delete` event.
- *Tombstone leakage in default / search reads* — `get_user_playbooks`, `search_user_playbooks`, and `count_user_playbooks_by_session` now exclude `MERGED`/`SUPERSEDED` when no explicit `status_filter` is supplied.

PR #191, opened 2026-06-21
Lineage Phase B3-pre — dedup write-side instrumentation + reconstruction (flag-OFF, non-destructive)

yilu331:
## What

Lineage Phase **B3-pre** — write-side instrumentation that makes the legacy `ProfileChangeLog` reconstructable from lineage, **plus** the read-side reconstruction + parity tooling (B3 Stage-1). **Ships flag-OFF and non-destructive** — it does NOT retire the legacy log (that's the future B3 retirement).

## Why

The B3 retirement plan was found (via `/review-design-doc` + a final review) to be infeasible as written: the profile-dedup path **hard-deletes** superseded profiles (no content, no linkage), so the content-free `lineage_event` log and the legacy `ProfileChangeLog` were **disjoint** — reconstruction couldn't reproduce the legacy log. This PR fixes the write side.

## Changes (behind a fail-CLOSED, default-OFF per-org flag)

- **`is_dedup_soft_delete_enabled`** — fail-closed flag (missing config → OFF; does NOT use the fail-open `is_feature_enabled`).
- **Dedup soft-delete** — when the flag is ON, the dedup path tombstones superseded profiles (`status=SUPERSEDED`, content retained) via `supersede_profiles_by_ids` and emits set-based `status_change` lineage under the run's shared `request_id`, atomically (no FTS/vec deletion; tombstones excluded from all reads by status filter). Flag OFF = byte-for-byte the current hard-delete.
- **`reconstruct_profile_change_log`** reworked to a **time-travel-stable column+event model**: `added` = the immutable `generated_from_request_id` column; `removed` = `status_change`/`superseded` events under the `request_id`. Dedup-scoped.
- **Parity gate test + one-shot parity script** + empty-`request_id` invariant.

## Safety

GC reclaims the tombstones this would create, so the flag **must not be enabled** until B2 GC has a retention story (PB-9/PB-5). The flag is default-OFF everywhere; the endpoint still serves the legacy table (a Stage-1 repoint was reverted). No table drop, no `mentioned_profiles` removal.

## Known follow-up (for the future B3 retirement, not this PR)

Add-only dedup runs emit no lineage event, so they're omitted from reconstruction (a `RECON-MISSING` the parity gate catches). The retirement must close this before any endpoint cutover.

Built via subagent-driven-development; full per-task + final whole-branch review (verdict: ready to ship flag-OFF).

## Summary by CodeRabbit

* **New Features**
  * Deduplicated profiles now follow a soft-delete (supersession) flow when the `dedup_soft_delete` feature flag is enabled.
  * Added bulk supersession support and lineage-based reconstruction for profile change logs.
  * Introduced a parity-check script to compare reconstructed vs legacy change-log outputs.
* **Tests**
  * Added/expanded integration tests for soft-delete behavior, lineage reconstruction edge cases, request_id invariants, and parity gate validation.
  * Added feature-flag unit coverage and API integration checks for `GET /api/profile_change_log`.

---
**Update (gap closed in-PR):** The add-only-run reconstruction completeness gap noted above is now **fixed in this PR** (T6) — the reconstruction discovers add-only dedup runs via a distinct `generated_from_request_id` union (OSS + enterprise), with tests. Reconstruction is complete (add-only runs included), not a deferred follow-up.

### DIFF
--- a/.superpowers/sdd/task-b3pre-1-report.md
+++ b/.superpowers/sdd/task-b3pre-1-report.md
@@ -1,0 +1,40 @@
+# Task B3-pre T1 Report: Dedup Soft-Delete + Set-Based Lineage
+
+## STATUS: COMPLETE
+
+## Summary
+
+Converted the profile dedup removal path from hard-delete to soft-delete behind `is_dedup_soft_delete_enabled` flag.
+
+## Changes
+
+### OSS (open_source/reflexio)
+
+**`reflexio/server/services/storage/storage_base/_profiles.py`**
+- Added abstract method `supersede_profiles_by_ids(user_id, profile_ids, request_id) -> int`
+
+**`reflexio/server/services/storage/sqlite_storage/_profiles.py`**
+- Added `supersede_profiles_by_ids` implementation: user_id-scoped, eligible = NULL|PENDING, one `conn.commit()` per call, `status_change` lineage event per updated id with shared `request_id`, guarded on `rowcount > 0`. No FTS/vec deletion.
+
+**`reflexio/server/services/profile/profile_generation_service.py`**
+- Branched `_finalize_extracted_items`: flag ON + non-empty `request_id` → `supersede_profiles_by_ids`; flag OFF or empty `request_id` → existing `delete_user_profile` loop (byte-for-byte unchanged).
+
+**`tests/server/services/profile/test_dedup_soft_delete_integration.py`** (new)
+- 13 tests: storage-level (8) + service-level (5). All pass.
+
+### Enterprise (reflexio_ext)
+
+**`reflexio_ext/server/services/storage/supabase_storage/_profiles.py`**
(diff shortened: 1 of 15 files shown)


yilu331:
CodeRabbit findings addressed in follow-up PR #193 (commit `22d2540`): negative-`limit` guard, `enabled_org_ids or []`, parity-script fail-closed-on-dup + INCONCLUSIVE-on-truncation, test hygiene (vacuous-MATCH guard, RECON/LEGACY label, exact set equality, empty-id seeding), ASCII `union`.
**Not changed (with reason):** the two "`get_profiles_by_generated_from_request_id` not org-scoped" findings — SQLite `profiles` has no `org_id` column (isolation is per-DB-file; enterprise per-schema), and the reconstruction's event pool is already org-scoped via `get_lineage_events`. A profiles-level org filter needs a schema migration — tracked as a separate follow-up, not a quick fix.

PR #197, opened 2026-06-21
Lineage B3d/B3e/B3f (OSS) — delete-path atomicity + #195/#196 CodeRabbit follow-ups

yilu331:
## Lineage B3d/B3e/B3f (OSS) — delete-path atomicity + CodeRabbit follow-ups from #195/#196

OSS/SQLite hardening from the B3b review-loop. Interactions/profiles emit no audit events for the search-index parts; the lineage emit on profile hard-deletes is preserved (audit-only-on-commit). Enterprise/Postgres deletes are transactional and unaffected.

### B3d — interaction public deletes (`sqlite_storage/_profiles.py`)
The 4 interaction-delete methods now clean `interactions_fts` + `interactions_vec` + the row in one atomic `with self._lock:`/single commit (raw DELETEs, vec guarded on `_has_sqlite_vec`). Closes a crash-window search-index drift AND an `interactions_vec` orphaned-vector leak (vec was never cleaned on delete).

### B3e — retention cascade (`sqlite_storage/_base.py`)
`_delete_interaction_search_rows` and `_delete_profile_search_rows` no longer self-commit (raw non-committing `_delete_in_chunks`), so the retention critical section (`_retention_perform_delete`) is truly atomic. `_delete_playbook_search_rows` left untouched (dual-use; 8 after-commit callers rely on its self-commit) — tracked follow-up.

### B3f — CodeRabbit findings that merged unaddressed in #195/#196
- **#196 (Major) — rowid-reuse race:** B3c cleaned profiles_fts/vec AFTER commit/outside the lock; `profiles` uses an implicit reusable rowid, so a concurrent insert could reuse it and the cleanup would delete the new row's vec entry. The 4 profile delete methods are now atomic (fts+vec+row+emit in one lock+commit, raw deletes) — race closed. Added the missing `profiles_vec` wipe to `delete_all_profiles`. (Playbooks/interactions are AUTOINCREMENT → no reuse → not affected.)
- **#195 (Major) — fail-closed validation:** `_is_fail_closed_flag_enabled` now uses `enabled is True` (strict; `{"enabled":"false"}`/`{"enabled":1}` no longer enable) and `isinstance(enabled_org_ids, list)` (a string no longer substring-matches). The fail-OPEN `is_feature_enabled` is intentionally unchanged.

### Tests
B3d (15) + B3e retention (3) + B3f profile-atomic (16) + flag strict-validation (9) + existing suites — all green (495 storage + flag tests). Vec assertions skip when sqlite-vec is unavailable (CI-env limitation).

### Process note
#195/#196 were merged before fully triaging CodeRabbit; B3f closes the two valid Major comments that slipped. The dual-use `_delete_playbook_search_rows` retention atomicity is the one remaining tracked item.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleted profiles and interactions are now reliably removed from search indexes and embeddings, preventing deleted items from appearing in search results.
  * Improved cleanup of historical data during retention operations to prevent orphaned search index entries.

* **Tests**
  * Added comprehensive integration tests validating atomic deletion of profiles and interactions with search index cleanup.
  * Enhanced feature flag validation tests for improved reliability.

### DIFF
--- a/reflexio/server/services/storage/sqlite_storage/_base.py
+++ b/reflexio/server/services/storage/sqlite_storage/_base.py
@@ -832,22 +832,35 @@ class SQLiteStorageBase(RetentionMixin, BaseStorage):
         self._delete_in_chunks("interactions", "request_id", request_ids)
 
     def _delete_interaction_search_rows(self, interaction_ids: list[int]) -> None:
+        """Remove fts + vec index rows for the given interaction IDs.
+
+        Non-committing: participates in the caller's transaction.  Only called
+        from inside the retention atomic block (_retention_perform_delete).
+        """
         if not interaction_ids:
             return
         self._delete_in_chunks("interactions_fts", "rowid", interaction_ids)
-        for interaction_id in interaction_ids:
-            self._vec_delete("interactions_vec", interaction_id)
+        if self._has_sqlite_vec:
+            self._delete_in_chunks("interactions_vec", "rowid", interaction_ids)
 
     def _delete_profile_search_rows(self, profile_ids: list[str]) -> None:
+        """Remove fts + vec index rows for the given profile IDs.
+
+        Non-committing: participates in the caller's transaction.  Only called
+        from inside the retention atomic block (_retention_perform_delete).
+        profiles_fts is keyed by profile_id (TEXT); profiles_vec by rowid (INT).
+        """
         if not profile_ids:
             return
-        rows = self._select_in_chunks(
-            "SELECT rowid, profile_id FROM profiles WHERE profile_id IN ({placeholders})",
-            profile_ids,
-        )
-        for row in rows:
-            self._fts_delete_profile(row["profile_id"])
-            self._vec_delete("profiles_vec", row["rowid"])
+        self._delete_in_chunks("profiles_fts", "profile_id", profile_ids)
+        if self._has_sqlite_vec:
+            rows = self._select_in_chunks(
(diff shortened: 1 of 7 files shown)


yilu331:
**CodeRabbit findings addressed in `e7cca4c` (B3g):**

1. **Cross-user sidecar deletes** (`delete_user_interaction`, + same class in `delete_user_profile`) — added a user-scoped existence check at the top of the lock; the method returns before any fts/vec/row delete if the id isn't owned by the requesting user. Prevents u1 from stripping u2's search/vec sidecars. (6 new cross-user tests, non-vacuous.)
2. **`delete_all_interactions_for_user` snapshot race** — moved the id-snapshot SELECT inside the `with self._lock:` block (via `conn.execute(...).fetchall()`, not `_fetchall`), so snapshot + deletes are one critical section. Same applied to `delete_oldest_interactions`.
3. **Bulk `IN (...)` deletes (param limits / S608)** — the new fts/vec sidecar deletes now use the non-committing `_delete_in_chunks` across `delete_all_interactions_for_user`, `delete_oldest_interactions`, `delete_all_profiles_for_user`, `delete_all_profiles_by_status`, `delete_profiles_by_ids`. Atomicity preserved (one lock + single commit).

All five delete methods keep one `with self._lock:` + single `conn.commit()`, rowcount-guarded lineage emit, user_id scoping, and return values. Playbook methods untouched.

PR #216, opened 2026-06-23
feat(llm): provider-safe schema for discriminated-union outputs (Option C)

yilu331:
## What

Make discriminated-union LLM output schemas **provider-safe by construction** — implements **Option C** from the design doc.

A Pydantic discriminated union (`Field(discriminator="kind")`) serializes to JSON Schema as `oneOf` + `discriminator`, which strict structured-output endpoints (OpenAI, minimax) reject with a 400. This is the root cause of Sentry **PYTHON-FASTAPI-9J**.

The shipped fix (#204) folds `oneOf` → `anyOf` inside `make_strict_json_schema`, **but only behind a provider gate** (`litellm.supports_response_schema` + an allowlist). That gate is exactly what failed in 9J (minimax under-reported `supports_response_schema=False`), and it leaves a standing dependency: the next under-reported provider re-breaks the same way.

## How

- `ProviderSafeUnionMixin` (in `server/llm/llm_utils.py`) overrides `__get_pydantic_json_schema__` to fold `oneOf` → `anyOf` and drop `discriminator` **at the model boundary**, via the shared `_fold_oneof_to_anyof` helper. Every caller of `model_json_schema()` — litellm, instructor, our own path — now gets a provider-safe schema **unconditionally**, with no dependency on provider detection.
- Applied to `PlaybookConsolidationOutput` (verified **internal-only** — not in the client, API schemas, or public docs).
- Only the **JSON (wire)** schema is rewritten; the **core (validation)** schema keeps the discriminator, so keyed dispatch and precise per-variant errors are preserved.

This is additive defense-in-depth — it does **not** remove #204's gate/normalizer (that path still handles strict-mode field stripping); it makes the oneOf-rejection class impossible at the source.

## Tests

`tests/server/services/playbook/test_consolidation_output.py`:
- `test_output_json_schema_is_provider_safe_by_construction` — `model_json_schema()` natively has no `oneOf`/`discriminator` (and keeps `anyOf`), with **no** `make_strict` and **no** provider gate.
- `test_provider_safe_schema_keeps_keyed_validation_errors` — a malformed `unify` payload still yields a single pinpointed error at the missing field (Option C's advantage over flattening the union).

Verified locally: ruff (lint+format) clean, pyright 0 errors, full consolidation-output suite (28) + mock-schema-compliance + litellm-unit (226 passed, 2 skipped) green.

## Design doc

`reflexio-enterprise` → `docs/superpowers/specs/2026-06-23-structured-output-provider-safe-schema-design.md` (PR #379), which covers Options A/B/C, the OpenAI `anyOf` first-key caveat, and the spike results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON schema compatibility with external API providers by adjusting union type representation.
  * Enhanced validation error messages to provide clearer feedback on union variant validation failures.

### DIFF
--- a/reflexio/server/llm/llm_utils.py
+++ b/reflexio/server/llm/llm_utils.py
@@ -4,7 +4,9 @@ import os
 from copy import deepcopy
 from typing import Any
 
-from pydantic import BaseModel
+from pydantic import BaseModel, GetJsonSchemaHandler
+from pydantic.json_schema import JsonSchemaValue
+from pydantic_core import CoreSchema
 
 
 def positive_int_env(name: str, default: int, logger: logging.Logger) -> int:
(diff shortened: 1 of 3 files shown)


yilu331:
@coderabbitai review

yilu331:
@coderabbitai full review

PR #219, opened 2026-06-24
feat(lineage): purge_content — content-only erasure that keeps lineage chains

yilu331:
Implements the `purge_content` primitive (specified but never built in the lineage design §5.1E/§14). Pairs with the enterprise PR that adds the Supabase/Postgres RPC + bumps this submodule pointer.

## What it does
GDPR erasure (`clear_user_data`) now **content-purges** a user's *referenced* lineage tombstones instead of hard-deleting them — blanks all PII (incl. `user_id`), keeps the `{id, status, merged_into, superseded_by, retired_at}` skeleton, emits an `op=purge` event — so `resolve_current` chains stay resolvable (`is_purged=True`) while the PII is gone. Standalone rows + interactions/requests hard-delete as before.

## Changes (TDD, per-task reviewed)
- **`has_inbound_lineage_refs`** (org-scoped, **not** user-scoped — cross-user chains).
- **`purge_content(entity_type, id)`** — blank = complement of the keep-allowlist; `profiles.user_id`→`''` (NOT NULL), `user_playbooks.user_id`→`NULL`; atomic blank+event in one commit (rowcount-guarded), FTS/vec cleanup after commit (the codebase's emit-before-commit hazard avoided); deterministic `request_id="purge_"+id` → idempotent (no duplicate event); PII-free event. Crash-window test proves no phantom event.
- **Both `clear_user_data` bodies** (SQLite override + BaseStorage default for Supabase/Postgres) gain the purge-vs-delete decision (tombstone OR pointed-to → purge; else hard-delete) via a shared `_partition_purge_vs_delete` helper. Erasure enumerates **all statuses incl. tombstones** (SUPERSEDED/MERGED).
- **`is_purged` contract** strengthened on `resolve_current`/`RecordRef` + a guard test. Audit found zero consumers dereference resolved content today (so live-pointed-to-survivor purge is safe now; the consumer-skip is a documented gated dependency for any future consumer).

## Verification
SQLite purge suite 10 passed; broader storage suite 527 passed; ruff + pyright clean. A final whole-branch review verified atomicity, blank-completeness parity with Supabase, tombstone-erasure completeness, and idempotency.

> **Governance:** the retained skeleton is "put beyond use" / Art. 18 restriction — **not** anonymisation. See the enterprise PR's DPO decision record.

## Summary by CodeRabbit

* **Documentation**
  * Updated record reference documentation to clarify handling when content is erased for privacy.
  * Enhanced function contract documentation describing how purged records should be treated.

* **New Features**
  * Content erasure for user data deletion now intelligently preserves lineage relationships while blanking personal information.

* **Tests**
  * Added comprehensive integration tests validating content erasure behavior and privacy compliance scenarios.

### DIFF
--- a/reflexio/models/api_schema/domain/entities.py
+++ b/reflexio/models/api_schema/domain/entities.py
@@ -468,7 +468,16 @@ class LineageContext(BaseModel):
 
 
 class RecordRef(BaseModel):
-    """Result of resolve_current — the live survivor's id and whether its body was purged."""
+    """Result of resolve_current — the live survivor's id and whether its body was purged.
+
+    Attributes:
+        id: Primary key of the live survivor record.
+        is_purged: True when the survivor's content body has been blanked by
+            ``purge_content`` (GDPR/erasure).  Any consumer that dereferences the
+            resolved record's content MUST treat ``is_purged=True`` as "erased —
+            skip or treat as absent."  Reading blank content as if it were valid
+            is a silent data-quality bug.
+    """
 
     id: str
     is_purged: bool = False
--- a/reflexio/server/services/lineage/resolve.py
+++ b/reflexio/server/services/lineage/resolve.py
@@ -24,6 +24,12 @@ def resolve_current(
 ) -> RecordRef | None:
     """Follow merged_into/superseded_by pointers to the live survivor.
 
+    **Consumer contract:** if the returned ``RecordRef.is_purged`` is ``True``,
+    the survivor's content body has been blanked by ``purge_content`` (GDPR/erasure).
+    Any consumer that dereferences the resolved record's content MUST skip or treat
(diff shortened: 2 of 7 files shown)


yilu331:
Triaged the 4 re-raised CodeRabbit comments against the current head:
- **source-window cleanup** (`_base.py`) — already fixed: `_delete_source_windows_for_user_playbook_ids(delete_upb_ids)` runs before the user_playbooks hard-delete.
- **is_purged assertion** (test) — already present: `assert ref.is_purged is True` in the cross-user test.
- **agent_playbook contract** — intentional documented limitation: `agent_playbook` has no `user_id` (out of scope), docstring states it raises `ValueError`.
- **hold lock through purge** — fixed in 4dcd03c: the purge loop now runs inside the same critical section as the hard-delete commit (continuous RLock hold).

PR #259, opened 2026-06-30
refactor(lineage-gc): split scheduler to tombstone-only (Phase 2)

guangyu-reflexio:
## Phase 2 — Premium reclaimer split (OSS side)

Reduces the shared lineage GC daemon to **tombstone-GC only** (core local data hygiene), splitting the premium **governance-retention** reclaimer out to an enterprise `Capability` (see the enterprise PR).

### What changed
- **`LineageGCScheduler` / `maybe_start_lineage_gc` reduced to tombstone-only** — dropped the `gc_governance_retention` branch, the `_is_governance_retention_enabled` helper, and the `GovernanceRetentionConfig` import; gate is now `if not cfg.lineage_gc.enabled`. Tombstone GC is unchanged (same entity loop, cutoff, high-volume tripwire, per-org resilience) and still default-on in **all** modes — no tombstone/PII-retention regression.
- **Dead-knob warning** — an OSS-only deployment that sets `audit_events_retention_enabled=True` now gets a one-time startup warning (the reclaimer lives in enterprise); the config field is documented enterprise-only.
- **`gc_governance_retention` stays an OSS abstract storage method** (consumed by the enterprise capability via the public `ctx.storage` interface) + a new backend-agnostic cross-backend contract test.

### Tests
- Effect-level shed test: the reduced scheduler **never** calls `gc_governance_retention` even with retention enabled, while tombstone GC still runs.
- Legacy-config path + `LineageGCConfig().enabled is True` default-flip tripwire.
- Backend-agnostic `gc_governance_retention` contract test (SQLite via the shared fixture; auto-covers Supabase when that param is added).

Behavior-preserving for tombstone GC; design doc reviewed across 3 rounds (incl. a focused verification that caught + closed a managed-tenant silent-skip risk).

## Summary by CodeRabbit

* **Bug Fixes**
  * Lineage garbage collection now focuses on removing only expired tombstones, with clearer behavior when retention options are unset or disabled.
  * Enterprise-only retention settings are treated as inactive in unsupported deployments, and a startup warning is shown to prevent confusion about “no-op” retention.
* **Tests**
  * Expanded unit and integration coverage for tombstone cleanup, governance-retention delete behavior (including batch limits and idempotency), and “dead-knob” warning scenarios.


guangyu-reflexio:
@coderabbitai review

PR #255, opened 2026-06-29
refactor: public OSS surface + Capability registry foundation (Phase 0 + 1A)

guangyu-reflexio:
## What

OSS-side of the boundary refactor, now covering **Phase 0 + Plan 1A** (paired with enterprise PR ReflexioAI/reflexio-enterprise#474).

### Phase 0 — public surface (shipped earlier)
- Promote `reflexio.server._auth` → public `reflexio.server.auth`.
- Extract governance validation helpers → public `reflexio.server.services.storage.governance_validation` (imported by SQLite + Supabase backends).
- Add public `Reflexio.get_storage()`.
Pure seam-hardening; no behavior change; no public client/API-schema change.

### Plan 1A — Capability registry foundation (new)
- New `reflexio.server.extensions`: process-global typed `ServiceKey` registry (`register_service`/`get_service`/`require_service`, de-dup); `Capability` ABC (no-op defaults); `HookRegistry` (facade over the existing hook setters); `AppContext`; `CapabilityRegistry`.
- `create_app(capabilities=)`: dual-path with the legacy params; **fail-loud** capability lifecycle (a raising `on_startup` aborts boot, cleaning up exactly the already-started capabilities); applies the configurator swap + billing-gate DI; threads the deployment role to `routers(role)`.

## Verification
- 15 OSS foundation tests (registry contract incl. de-dup/require; `create_app` dual-path, fail-loud partial-cleanup invariant, legacy-path unchanged); 203 existing server tests pass; `import reflexio` standalone (boundary guard). ruff + pyright clean.

## Scope
Phase 0 + Plan 1A foundation. Runtime-service consumers (redaction) land in a later phase. Client surface, API schemas, and the core learning loop are untouched.

## Summary by CodeRabbit

* **New Features**
  * Added enterprise app “capabilities” for capability-provided routing plus async startup/shutdown lifecycle hooks.
  * Introduced a new public `get_storage()` method to retrieve the configured storage backend.
  * Enabled aggregation prompt processing via service-key registration.
* **Bug Fixes**
  * Centralized and strengthened governance validation/canonicalization for storage audit/erase handling.
  * Improved capability lifecycle handling so shutdown continues even if a capability’s shutdown fails.
* **Tests**
  * Expanded coverage for capabilities wiring (including router mount guard), runtime service registry, aggregation prompt processor injection, storage getter behavior, and governance ref validation.

PR #266, opened 2026-07-01
feat: surface OSS structured logs in dashboard

wenchanghan:
## Summary
- Adds a lean OSS structured-log path so local operators can see recent Reflexio warning/error/critical events from the dashboard and `GET /api/logs`.
- Stores structured log events in the same OSS SQLite DB as local Reflexio data, with bounded row retention, capped text fields, literal search semantics, and best-effort fail-open writes.
- Adds a first-class `/logs` dashboard page with level filters, time window, search, manual refresh, and newest/oldest sorting.
- Moves Configure and Logs into an Operations sidebar section so operator pages are separated from domain/API resources.
- Adds `GET /api/logs` to the API reference under Logging, separate from the `/logs` operator page.

## Changes
- Backend
  - Adds `reflexio/server/structured_logging.py` with a SQLite store, logging handler, lifecycle handle, OSS DB-path resolution, and retention enforcement.
  - Installs the structured logging handler during data-plane startup and closes it during shutdown or failed startup cleanup.
  - Captures first-party `reflexio.*` warning/error/critical log records to avoid third-party warning noise filling the operator window.
- API
  - Adds `GET /api/logs` with `levels`, `since`, `q`, and `limit` filters.
  - Keeps the route behind the existing org dependency path when auth is configured.
  - Bounds search input and escapes SQL LIKE wildcards so `%` and `_` are treated as literal search text.
- Dashboard
  - Adds `/logs` as an operator page.
  - Shows timestamp, level, message, and traceback text when present.
  - Adds multi-level filtering, time-window filtering, search, manual refresh, and newest/oldest sorting.
  - Adds Operations sidebar grouping for Configure and Logs.
- Docs and tests
  - Documents the operator log page and adds an API reference page for `GET /api/logs`.
  - Splits operations and admin endpoint buckets in the server README.
  - Adds coverage for log capture, API filtering and validation, auth dependency behavior, startup cleanup, retention/capping, fail-open writes, lock contention, literal search, and first-party logger scoping.

## Diagrams

```mermaid
sequenceDiagram
  participant Startup as Reflexio data-plane startup
  participant Config as OSS configurator
  participant DB as Same OSS SQLite DB file
  participant Domain as Existing domain tables
  participant LogTable as structured_log_events table
  participant UI as /logs dashboard

  Startup->>Config: resolve active StorageConfigSQLite.db_path
  Config-->>Startup: local SQLite path used by Reflexio storage
  Startup->>DB: install structured log handler against same DB file
  Note over DB: Shared physical DB file and connection target
  Domain->>DB: existing profile/playbook/session storage continues unchanged
  Note over Domain,LogTable: Domain tables and log table are separate schemas
  Startup->>LogTable: create structured_log_events table and indexes if missing
  LogTable-->>UI: GET /api/logs reads only structured_log_events
  Note over UI,Domain: Logs UI/API do not query or mutate domain tables
```

## Screenshots

**Before**

Left nav before operator pages were separated from API/domain docs:

Before left nav

**After**

Default Logs view with Error/Critical rows and traceback text:

After logs default

Warning filter included:

After warning filter

Literal search for `%` and `_`:

After literal search

Oldest-first sorting:

After oldest-first sort

Logging API docs entry in the top API navigation:

After logging API docs

API failure state with stale rows cleared:

After API failure

Mobile left-nav grouping:

After mobile nav

## Test Plan
- `uv run ruff check reflexio/server/structured_logging.py reflexio/server/api_endpoints/logs_api.py reflexio/models/api_schema/logs_schema.py reflexio/server/api.py tests/server/test_structured_logging.py tests/server/services/test_configurator.py`
- `uv run pyright reflexio/server/structured_logging.py reflexio/server/api_endpoints/logs_api.py reflexio/models/api_schema/logs_schema.py tests/server/test_structured_logging.py tests/server/services/test_configurator.py`
- `uv run pytest tests/server/test_structured_logging.py tests/server/services/test_configurator.py tests/server/test_create_app.py tests/server/test_create_app_capabilities.py tests/server/api_endpoints/test_api_routes.py tests/server/api_endpoints/test_health_api.py -n 0 --no-cov -q` -> 62 passed
- `cd docs && npm run lint && npx tsc --noEmit && npm run build` -> passed; existing lint warnings remain in unrelated docs files.
- Browser docs check: opened `http://localhost:8092/logging/get-logs` and verified the top API nav shows Logging -> Get Logs, while the lower Operations section still contains Configure and Logs.
- E2E validation on local services (`uv run reflexio services start --no-reload --workers 1`):
  - Confirmed the previous dashboard API failure was a docs-only smoke issue; with backend and docs running, `/logs` loaded through `http://localhost:8061/api/logs` without an API failure banner.
  - Triggered a live first-party backend warning and verified it was captured in SQLite and returned by `GET /api/logs`.
  - Emitted marked warning/error/critical rows, including traceback text, through the structured logging path and verified API default levels, explicit multi-level filters, literal `%`/`_` search, limit bounds, invalid levels, and max search length handling.
  - Verified in browser: default Error/Critical rows, Warning inclusion, Source logger column, search narrowing, literal percent search, oldest-first sorting with deterministic timestamps, traceback display, API failure state with stale rows cleared, mobile navigation grouping, and the sidebar order with API reference groups above Operations.
  - Removed marked validation rows from the local SQLite DB after screenshots were captured.
- Full repo baseline remains outside this PR:
  - `uv run ruff check .` and `uv run pyright` still report existing unrelated issues.
  - Full pytest previously showed unrelated benchmark/key, optimizer, and suite-order/vector-storage failures; focused changed-area and affected-slice tests pass after rebasing onto latest `main`.

## Summary by CodeRabbit

* **New Features**
  * Added a new Logs page in the dashboard to view recent warning, error, and critical events.
  * Added log filtering, search, sorting, and refresh controls for easier troubleshooting.
  * Added a new Logging section in the navigation with direct access to logs and configuration pages.

* **Bug Fixes**
  * Improved log retrieval reliability and validation for time ranges, log levels, and search input.
  * Preserved recent log history with bounded storage and safer handling of large messages.


wenchanghan:
### Code review

No high-confidence bugs found. Checked for bugs, CLAUDE.md compliance, git history context, prior-PR comment patterns, and code-comment invariants.

---

#### Observations for follow-up (below review threshold, surfaced for completeness)

**1. LIKE wildcards not escaped in search query** (confidence: 75/100)

`structured_logging.py` builds `like_filter = f"%{q}%"` without escaping `%` and `_` in the user's search term. Searching for a literal `%` (common in URL-encoded strings, format patterns) or `_` (common in snake_case identifiers in tracebacks) returns broader results than expected. Not a security issue (parameterized queries are used), but incorrect search behavior for edge-case inputs. Fix: escape with `q.replace("%", "\\%").replace("_", "\\_")` and add `ESCAPE '\'` to the SQL.

https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/reflexio/server/structured_logging.py#L138-L140

**2. Stale log rows persist after fetch error in dashboard** (confidence: 75/100)

In `logs-dashboard.tsx`, when `loadLogs` fails the `error` state is set but `items` is not cleared. The UI renders both the error banner and the previous successful response's rows, which could mislead an operator into thinking stale data is current. The codebase's own `use-execution.tsx` hook clears results on error. One-line fix: add `setItems([])` before `setError(...)` in both error branches.

https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/docs/components/logs/logs-dashboard.tsx#L74-L82

**3. Missing `# noqa: BLE001` on broad `except Exception` in `api.py`** (confidence: 50/100, from prior-PR pattern)

PR #255 established the convention of marking intentional broad `except Exception` blocks with `# noqa: BLE001` plus a rationale comment. The new `_start_data_plane_services` in `api.py` has a bare `except Exception:` that re-raises after logging and cleanup, but lacks the suppression comment. The handler in `structured_logging.py` already follows the convention correctly (`# noqa: BLE001 - logging must fail open`).

https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/reflexio/server/api.py#L3527-L3532

---

#### Simplification opportunities (no changes made)

**S1. `toggleLevel` in `logs-dashboard.tsx` can be simplified**

The current implementation uses two `LOG_LEVELS.filter(...)` calls to add/remove a level while preserving LOG_LEVELS ordering. This can be reduced to a straightforward toggle:
```ts
setSelectedLevels((cur) =>
  cur.includes(level)
    ? cur.length > 1 ? cur.filter((l) => l !== level) : cur
    : [...cur, level]
);
```
Order preservation isn't necessary since the levels are looked up by value, not position.

https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/docs/components/logs/logs-dashboard.tsx#L87-L100

**S2. `_VALID_LEVELS` is duplicated across `logs_api.py` and `structured_logging.py`**

Both files define `_VALID_LEVELS = frozenset({"warning", "error", "critical"})`. This could be a single constant imported from `logs_schema.py` (which already defines `LogLevel = Literal["warning", "error", "critical"]`) or from `structured_logging.py`.

https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/reflexio/server/api_endpoints/logs_api.py#L14
https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/reflexio/server/structured_logging.py#L36

**S3. 10-element params tuple in `StructuredLogStore.query` is fragile**

The query method builds a 10-element positional params tuple that must stay in sync with 10 `?` placeholders in the SQL string. A dict-based approach with named `:placeholders` would be self-documenting and less error-prone as the query evolves.

https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/reflexio/server/structured_logging.py#L130-L160

**S4. `getErrorMessage` does redundant type narrowing**

The function chains three `if` blocks that each re-check `payload && typeof payload === "object"`. A single guard clause at the top (`if (!payload || typeof payload !== 'object') return fallback`) would flatten the logic.

https://github.com/ReflexioAI/reflexio/blob/70dfc578eec25e21ae9ba9fb89a31ff7d896a895/docs/components/logs/logs-dashboard.tsx#L42-L63

reflexio/server/api_endpoints/logs_api.py
```diff
@@ -0,0 +1,105 @@
...
+_DEFAULT_LEVELS = {"error", "critical"}
+
+
+@router.get("/logs", response_model=GetLogsResponse)
+def get_logs(
+    request: Request,
+    _org_id: Annotated[str, Depends(default_get_org_id)],
+    levels: Annotated[str | None, Query()] = None,
+    since: Annotated[str | None, Query()] = None,
+    q: Annotated[str | None, Query()] = None,
```
wenchanghan: **🩺 Stability & Availability | 🟡 Minor | Suggestion**

`/api/logs` accepts an unbounded `q` string. An operator (or a misbehaving script) can send a multi-MB `q`, which will be interpolated into three `LIKE` patterns and force a full scan on every row that survives the level/since filters. The retention cap is 10k rows but each row is up to ~40 KB, so a pathological `q` can stall the dashboard query thread.

A small bounded length (e.g. `Query(max_length=256)`) keeps the substring search honest and matches the spirit of the `_cap` budget already enforced on `message` / `exception_text` at write time. It also matches the shape of `_parse_limit`'s explicit upper bound.

reflexio/server/api.py
```diff
@@ -3418,6 +3419,121 @@ def _wire_capabilities(
...
+    # drives a per-org worker with org-scoped claims, so it is not limited
+    # to the bootstrap org. The bootstrap org is only used to read config
+    # and to seed cross-org discovery.
+    bootstrap_org_id = _resolve_lifespan_org_id(get_org_id)
+    structured_log_db_path = resolve_oss_structured_log_db_path(bootstrap_org_id)
+    if structured_log_db_path is not None:
+        structured_logging_handle = install_structured_logging(structured_log_db_path)
+        app.state.structured_logging_handle = structured_logging_handle
+    else:
+        app.state.structured_logging_unavailable_reason = (
```
wenchanghan: **🧹 Dead Code | 🟡 Minor | Suggestion**

`app.state.structured_logging_unavailable_reason` is set here but never read anywhere in the tree (`grep -rn "structured_logging_unavailable_reason" reflexio/` only shows this single write site). The `/api/logs` endpoint just checks for the handle and falls through to a 503 with a generic message, so the per-app reason text never reaches an operator.

Two equally small options:

1. Drop the attribute and let the 503 detail stay generic.
2. Surface it — the endpoint can read `getattr(state, "structured_logging_unavailable_reason", None)` and return it as the 503 detail so the dashboard can render "Structured logs are unavailable: …" instead of a hard-coded string.

Right now it's the worst of both worlds: extra startup work, zero operator value.

reflexio/server/structured_logging.py
```diff
@@ -0,0 +1,327 @@
...
+                return StructuredLoggingHandle(handler)
+            raise RuntimeError(
+                "Reflexio structured logging is already installed for "
+                f"{handler.db_path}; refusing to route records to {db_path}"
+            )
+
+        store = StructuredLogStore(db_path, row_limit=row_limit)
+        handler = StructuredLogHandler(store)
+        handler._ref_count = 1
+        root_logger.addHandler(handler)
```
wenchanghan: **🩺 Stability & Availability | 🟡 Minor | Suggestion**

`StructuredLogHandler` is added to the root logger in `install_structured_logging`, so it captures WARNING+ from every logger in the process — including uvicorn, asyncio, httpx, openai, etc. For an "operator dashboard" sink that's mostly intended for first-party diagnostics, this can flood the 10k-row retention window with chatty third-party warnings (e.g. `httpx` "HTTP Request: …", `asyncio` "SelectorSocket … has no data") and push real app warnings out of the visible window.

Two reasonable shapers:

1. Filter at the handler — only emit rows for loggers whose name starts with `"reflexio."` (plus the structured log store's own writes, which are already keyed off `record.name`).
2. Attach the handler to a named logger subtree instead of root, e.g. `logging.getLogger("reflexio")`, so third-party noise is excluded by configuration.

Either way, the level/message filtering stays unchanged and the dashboard continues to show app diagnostics.

reflexio/server/structured_logging.py
```diff
@@ -0,0 +1,327 @@
...
+                    OR (? = 1 AND level = 'error')
+                    OR (? = 1 AND level = 'critical')
+                )
+                AND (? IS NULL OR timestamp >= ?)
+                AND (
+                    ? IS NULL
+                    OR message LIKE ?
+                    OR exception_text LIKE ?
+                    OR logger_name LIKE ?
+                )
```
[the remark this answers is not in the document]
wenchanghan: Agree on the fix — adding a small regression test alongside the existing `test_logs_api_search_since_iso_and_limit` would lock it in. Specifically, after escaping, assert:

- `q=100%` matches a row whose message literally contains `100%` and does **not** match a row that contains `100` plus other characters from unintended wildcards.
- `q=foo_bar` only matches a row literally containing `foo_bar`, not `foo` + any single char + `bar`.

A second invariant worth asserting: an unescaped `q` with embedded `%`/`_` still goes through the level/since filters intact (no truncation at the SQL level).

Happy to pair this with the `q` length cap suggested on `logs_api.py` so the search input is bounded both ways.

wenchanghan (commented):
## Code review — `Surface OSS structured logs in dashboard`

Solid v1 — the SQLite sink, `/api/logs` filter validation, lifespan cleanup, dashboard UX, and the configurator path extraction are all coherent and well-tested. The below threads capture the items worth a follow-up before merge:

- `_HANDLER_LOCK` + `_ref_count` ref-counting handles the same-DB re-use case nicely, but does nothing for the cross-DB conflict path in `install_structured_logging` — calling `raise RuntimeError` after the first install means a test or hot-reload that tries to swap the DB path crashes the process. Worth deciding whether that should be `RuntimeError` or a logged warning with a fallback to no-op.
- `StructuredLogStore` and `StructuredLoggingHandle` are tightly coupled by `handler._ref_count` access from outside the handler class. A tiny `_inc_refcount` / `_dec_refcount` helper on the handler would let `StructuredLoggingHandle.close()` stop reaching into a private attribute and keep the ref-count invariant in one place.
- The frontend `requestUrl` correctly memoizes per filter, but the `useEffect` that triggers `loadLogs` has `loadLogs` and `refreshNonce` as deps while `loadLogs` is itself memoized on `requestUrl` — clicking Refresh always works because `refreshNonce` bumps, but if `loadLogs` were ever swapped for an inline fetch the auto-refetch contract would silently break. A short comment on the useEffect would help the next reader.
- Default `q` strip is done twice (`if q and q.strip()` then `q.strip()`); cosmetic, but worth tidying while the file is open.

Everything else looked clean. Approving with the threads above as follow-ups.

reflexio/server/README.md
```diff
@@ -79,7 +80,7 @@ Description: FastAPI backend server that processes user interactions to generate
...
 - **Evaluation**: `POST /api/get_evaluation_overview`, `POST /api/evaluations/regenerate`, `GET /api/evaluations/regenerate/{job_id}`, `DELETE /api/evaluations/regenerate/{job_id}`, `POST /api/evaluations/grade_on_demand`, `GET /api/evaluations/shadow_comparisons/recent`
 - **Braintrust**: `POST /api/braintrust/connect`, `POST /api/braintrust/select_projects`, `GET /api/braintrust/status`, `DELETE /api/braintrust/connection`, `POST /api/braintrust/sync`
-- **Operations/admin**: `GET /api/get_operation_status`, `POST /api/cancel_operation`, `POST /api/admin/cache/invalidate`, `DELETE /api/delete_interaction`, `DELETE /api/delete_request`, `DELETE /api/delete_session`, `DELETE /api/delete_requests_by_ids`, `DELETE /api/delete_all_interactions`, `POST /api/clear_user_data`
+- **Operations/admin**: `GET /api/logs`, `GET /api/get_operation_status`, `POST /api/cancel_operation`, `POST /api/admin/cache/invalidate`, `DELETE /api/delete_interaction`, `DELETE /api/delete_request`, `DELETE /api/delete_session`, `DELETE /api/delete_requests_by_ids`, `DELETE /api/delete_all_interactions`, `POST /api/clear_user_data`
```
wenchanghan: it may be a good time to separate out admin & operation ones

Admin ones are for things like "delete interaction"
Operation ones are for things like "/logs", "/configs"

reflexio/server/README.md
```diff
@@ -80,7 +80,8 @@ Description: FastAPI backend server that processes user interactions to generate
...
 - **Evaluation**: `POST /api/get_evaluation_overview`, `POST /api/evaluations/regenerate`, `GET /api/evaluations/regenerate/{job_id}`, `DELETE /api/evaluations/regenerate/{job_id}`, `POST /api/evaluations/grade_on_demand`, `GET /api/evaluations/shadow_comparisons/recent`
 - **Braintrust**: `POST /api/braintrust/connect`, `POST /api/braintrust/select_projects`, `GET /api/braintrust/status`, `DELETE /api/braintrust/connection`, `POST /api/braintrust/sync`
-- **Operations/admin**: `GET /api/logs`, `GET /api/get_operation_status`, `POST /api/cancel_operation`, `POST /api/admin/cache/invalidate`, `DELETE /api/delete_interaction`, `DELETE /api/delete_request`, `DELETE /api/delete_session`, `DELETE /api/delete_requests_by_ids`, `DELETE /api/delete_all_interactions`, `POST /api/clear_user_data`
+- **Operations**: `GET /api/logs`, `GET /api/get_operation_status`, `POST /api/cancel_operation`
```
wenchanghan: are these domain- / playbook specific or something related to operating Reflexio OSS (for service running / failures)?

reflexio/server/README.md
```diff
@@ -80,7 +80,8 @@ Description: FastAPI backend server that processes user interactions to generate
...
+- **Admin**: `POST /api/admin/cache/invalidate`, `DELETE /api/delete_interaction`, `DELETE /api/delete_request`, `DELETE /api/delete_session`, `DELETE /api/delete_requests_by_ids`, `DELETE /api/delete_all_interactions`, `POST /api/clear_user_data`
```
wenchanghan: is this "cache invalidation" one related to operating the service (i.e. recovering from a service failure) than playbook- / domain-specific? If so, this would belong to operations, not admins

wenchanghan: holistically look at the grouping of these APIs -> really confirm what they are trying to do and group them correctly

operation-related: this is about ensuring reflexio service is healthy (or operations that could mitigate issues to bring the service back to healthy state). These are unrelated to learning / publish / search

admin-related: this is about managing playbook, etc.

docs/lib/constants.ts
```diff
@@ -7,5 +7,6 @@ export const NAV_GROUPS = [
   { id: "agent-evaluation", name: "Agent Evaluation", icon: "BarChart" },
   { id: "unified-search", name: "Unified Search", icon: "Search" },
   { id: "generation", name: "Generation", icon: "Cpu" },
-  { id: "configuration", name: "Configuration", icon: "Settings" },
+  { id: "operations", name: "Operations API", icon: "Activity" },
+  { id: "configuration", name: "Config API", icon: "Settings" },
```
wenchanghan: do we really need to rename this to Config API, now that this left-nav item is clearly in the API documentation left-nav area (top part), separating it out from "Operations - Configure"?

docs/components/layout/sidebar.tsx
```diff
@@ -53,30 +56,67 @@ export function Sidebar() {
...
         <div className="space-y-1">
           {resourceGroups.map((group) => (
-            <SidebarGroup
-              key={group.id}
-              group={group}
-              pathname={pathname}
-            />
+            <SidebarGroup key={group.id} group={group} pathname={pathname} />
           ))}
+          <div className="mt-4 border-t border-border pt-3">
```
wenchanghan, line 73: is it right to have the top part (API documentations for Interactions, User Profiles, etc.) to have a "resourceGroups.map" while the Operations section (including Configure and Logs) to be hardcoded without a "map/loop"?

docs/lib/methods/operations.ts
```diff
@@ -0,0 +1,45 @@
+import { MethodDef } from "../types";
+
+export const operationMethods: MethodDef[] = [
```
wenchanghan: only seeing logs-related ones under this generic docs/lib/methods/operations.ts
expected?

wenchanghan:
Pausing this change while we revisit the OSS operator diagnostics contract and the right customer experience for backend issue visibility. The implementation and validation notes here remain useful context, but this PR should not be merged until that design direction is settled.

PR #285, opened 2026-07-03
feat(lineage): register_per_org_sweep seam for per-org reclamation

guangyu-reflexio:
## Summary

Adds a **per-org** reclamation sweep seam to the lineage GC scheduler — the per-org counterpart to the merged `register_global_sweep` (global) seam (#283). A deployment can register a `Callable[[str, int], int]` that the scheduler invokes **once per org** inside its existing per-org tick loop, keeping the OSS scheduler config-agnostic (no enterprise imports).

The enterprise side uses this to fold its per-org `GovernanceRetentionScheduler` (audit-event retention) into the shared scheduler and delete a standalone daemon (see the enterprise PR).

## Changes

- `gc_scheduler.py`:
  - `_per_org_sweep_hooks` list + `register_per_org_sweep(fn: Callable[[str, int], int])` + `clear_per_org_sweeps()` (test hook), mirroring `register_global_sweep`.
  - `LineageGCScheduler._gc_tick`: after the Class-B block, inside the per-org loop, invokes each registered per-org sweep **unconditionally** with `(org_id, now)` and per-sweep failure isolation (`capture_anomaly("lineage.per_org_sweep.failed", org_id=, sweep=)`). The gate is the enterprise closure's own per-org config — not `lineage_gc`/`expiry_reclamation`.
  - `maybe_start_lineage_gc`: start-gate widened to also start when **any** reclamation sweep is registered (`_per_org_sweep_hooks`/`_global_sweep_hooks` non-empty), so a per-org reclaimer whose real gate lives in a non-bootstrap tenant's config is never silently skipped (preserving the folded daemon's unconditional-start invariant).
  - Docstring: per-org sweeps documented; stale `GovernanceRetentionCapability` references replaced with the generic per-org-sweep description.

## Tests

`test_gc_scheduler.py`: per-org sweep runs once per org; failure-isolation (anomaly + siblings continue); `maybe_start` starts when a per-org sweep is registered even with all config flags off; existing `test_gc_tick_never_runs_governance_retention` stays green (empty hooks → OSS never runs governance). 46 lineage tests pass.

With no registered sweeps (OSS default), behavior is byte-for-byte identical to before.

## Related PRs
- open_source/reflexio: #283 (the global-sweep seam this parallels)

## Summary by CodeRabbit

* **New Features**
  * Added support for per-organization retention sweeps, allowing scheduled maintenance to run for each org.
  * The scheduler now starts when retention-related sweeps are registered, even if other GC settings are off.

* **Bug Fixes**
  * Improved sweep reliability so one failing org-level sweep does not stop other sweeps from running.
  * Updated retention-related warnings and behavior to better reflect enterprise sweep handling.

### DIFF
--- a/reflexio/models/config_schema.py
+++ b/reflexio/models/config_schema.py
@@ -725,9 +725,10 @@ class ExpiryReclamationConfig(BaseModel):
 
 class GovernanceRetentionConfig(BaseModel):
     """Audit-event retention policy. **Enterprise-only:** reclamation is performed
-    by the reflexio_ext GovernanceRetentionCapability. In an OSS-only deployment
-    these knobs are accepted but inert (the OSS lineage scheduler does not reclaim
-    audit events) and the server logs a startup warning when retention is enabled.
+    by an enterprise per-org reclamation sweep registered via
+    ``register_per_org_sweep``. In an OSS-only deployment these knobs are accepted
+    but inert (the OSS lineage scheduler does not reclaim audit events) and the
+    server logs a startup warning when retention is enabled.
     """
 
     audit_events_retention_enabled: bool = False
--- a/reflexio/server/services/lineage/gc_scheduler.py
+++ b/reflexio/server/services/lineage/gc_scheduler.py
@@ -411,10 +411,11 @@ def maybe_start_lineage_gc(
         cfg = ctx.configurator.get_config()
 
         # Dead-knob warning: audit-event retention is an ENTERPRISE-only feature
-        # (handled by reflexio_ext GovernanceRetentionCapability). In an OSS-only
-        # deployment the knob is accepted but does nothing — warn loudly. Detected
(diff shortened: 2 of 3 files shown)


guangyu-reflexio:
Addressed all 3 nitpicks in `ae28e197`:
- Synced the `maybe_start_lineage_gc` docstring with the widened start gate (governance flag + registered hooks).
- Per-org sweep loop now computes `now` once and logs the deleted count (`event=per_org_sweep ... deleted=N`), matching the Class-B and global-sweep paths.
- On the mutation-safety note: the per-org loop already uses the identical direct-iteration as `_global_sweep_hooks`, and hooks are registered pre-lifespan (never mutated during ticks; `clear_*` is test-only), so there's no runtime race — and `ruff` PERF101 forbids the `list()` snapshot. Extracted the loop into `_run_per_org_sweeps` to keep `_gc_tick` under the complexity limit.

@coderabbitai review

PR #287, opened 2026-07-04
fix(llm,storage): harden Windows extraction paths

wenchanghan:
## Why
Windows local extraction can fail before Reflexio gets a usable model response:
- Windows code pages can break non-ASCII subprocess input unless the local CLI bridge is pinned to UTF-8.
- Large extraction schemas can exceed Windows command-line length when sent through `--append-system-prompt`.
- Extensionless CLI override paths can miss executable shims commonly installed on Windows.
- Some storage callers can hand timestamp values that need safe normalization without breaking existing open-bound sentinels.

## What Changed
- Local CLI provider
  - Pins Claude and Codex subprocess text encoding to UTF-8.
  - Uses replacement decoding only on Windows; non-Windows stays strict.
  - Keeps small Claude prompts on `--append-system-prompt`.
  - Moves only large Windows Claude system prompts into stdin, without Codex-specific `## Task` framing.
  - Resolves extensionless `CLAUDE_SMART_CLI_PATH` overrides through `PATHEXT` plus `.cmd`/`.exe`/`.bat` Windows shims.
  - Skips `.ps1` shim auto-resolution because PowerShell scripts require an explicit PowerShell host.
- SQLite timestamp helper
  - Normalizes contemporary millisecond epochs to seconds before ISO conversion.
  - Preserves documented open-bound sentinel behavior such as `to_ts=10**12`.
- Tests
  - Covers non-ASCII subprocess input, Windows large prompt fallback, large structured-output schemas, `.cmd`/`.exe`/`.bat` override shims, `.ps1` exclusion, and seconds/milliseconds/sentinel timestamp cases.

## Compatibility Notes
- Public API timestamp contracts remain Unix seconds; no public docs update is needed for this PR.
- The timestamp tolerance is intentionally bounded so historical open-bound sentinels remain in seconds-space.
- The Windows prompt fallback is only used when the system prompt is large enough to risk argv limits.

## Validation
- `uv run pytest --no-cov tests/server/llm/test_claude_code_provider.py tests/server/services/storage/test_sqlite_storage.py tests/server/api_endpoints/test_evaluation_overview_api.py -q` - 112 passed.
- `uv run pytest --no-cov tests/lib/test_interactions_unit.py tests/e2e_tests/test_playbook_workflows.py -q` - 39 passed, 16 skipped.
- `uv run ruff check reflexio/server/llm/providers/claude_code_provider.py reflexio/server/services/storage/sqlite_storage/_base.py tests/server/llm/test_claude_code_provider.py tests/server/services/storage/test_sqlite_storage.py tests/server/api_endpoints/test_evaluation_overview_api.py` - passed.
- `uv run ruff format --check reflexio/server/llm/providers/claude_code_provider.py reflexio/server/services/storage/sqlite_storage/_base.py tests/server/llm/test_claude_code_provider.py tests/server/services/storage/test_sqlite_storage.py tests/server/api_endpoints/test_evaluation_overview_api.py` - passed.
- `uv run pyright reflexio/server/llm/providers/claude_code_provider.py reflexio/server/services/storage/sqlite_storage/_base.py tests/server/llm/test_claude_code_provider.py tests/server/services/storage/test_sqlite_storage.py tests/server/api_endpoints/test_evaluation_overview_api.py` - passed.
- `git diff --check` - passed.
- Earlier full `uv run pytest -q` was attempted on this branch: coverage passed at 81.78%, but unrelated benchmark/eval/optimizer tests failed locally because `OPENAI_API_KEY` is not set.

### DIFF
--- a/reflexio/server/llm/providers/claude_code_provider.py
+++ b/reflexio/server/llm/providers/claude_code_provider.py
@@ -70,6 +70,7 @@ _CODEX_COMPAT_SCRIPT_NAMES = (
 _CODEX_COMPAT_SCRIPT_NAME_SET = set(_CODEX_COMPAT_SCRIPT_NAMES)
 _DEFAULT_TIMEOUT_SECONDS = 120
 _DEFAULT_CLI_MODEL = "claude-sonnet-5"
+_WINDOWS_ARGV_SYSTEM_PROMPT_LIMIT = 3_000
 
 _TRUTHY_ENV_VALUES = {"1", "true", "yes"}
 _UNSUPPORTED_PARAMS_WARNED: set[str] = set()
@@ -116,6 +117,18 @@ def _candidate_codex_compat_path() -> Path | None:
     return None
 
 
+def _windows_executable_cli_path(cli_path: str) -> str:
+    if os.name != "nt":
+        return cli_path
+    filename = cli_path.rsplit("/", 1)[-1].rsplit("\\", 1)[-1]
+    if "." in filename:
+        return cli_path
+    cmd_path = f"{cli_path}.cmd"
+    if os.path.exists(cmd_path):  # noqa: PTH110 - Path() follows patched os.name.
+        return cmd_path
+    return cli_path
+
+
 def _resolve_cli_path() -> str | None:
     """Return the path to the active host CLI, or None if unavailable.
 
@@ -129,9 +142,9 @@ def _resolve_cli_path() -> str | None:
     """
     override = os.environ.get(_ENV_CLI_PATH)
     if override:
-        candidate = Path(override)
-        if candidate.is_file() and os.access(candidate, os.X_OK):
-            return str(candidate)
+        candidate = _windows_executable_cli_path(override)
+        if os.path.isfile(candidate) and os.access(candidate, os.X_OK):  # noqa: PTH113
+            return candidate
         _LOGGER.warning(
             "%s=%s is not an executable file; falling back to PATH",
             _ENV_CLI_PATH,
@@ -432,6 +445,7 @@ def _run_claude_stream(
 ) -> ParseResult:
     """Invoke ``claude -p --output-format stream-json`` and return a ParseResult."""
     model = os.environ.get(_ENV_MODEL) or _DEFAULT_CLI_MODEL
+    inline_system_prompt = _should_inline_system_prompt(system_prompt)
     cmd = [
         cli_path,
         "-p",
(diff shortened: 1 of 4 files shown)


reflexio/server/llm/providers/claude_code_provider.py
```diff
@@ -116,6 +117,18 @@ def _candidate_codex_compat_path() -> Path | None:
     return None
 
 
+def _windows_executable_cli_path(cli_path: str) -> str:
```
wenchanghan: this method is not invoked only for windows flow, right? if that's the case, do not use windows in name

reflexio/server/llm/providers/claude_code_provider.py
```diff
@@ -116,6 +117,18 @@ def _candidate_codex_compat_path() -> Path | None:
...
+    if os.name != "nt":
+        return cli_path
+    filename = cli_path.rsplit("/", 1)[-1].rsplit("\\", 1)[-1]
```
wenchanghan: please add comment - it's not apparent why this split is needed

reflexio/server/llm/providers/claude_code_provider.py
```diff
@@ -550,6 +573,10 @@ def _codex_prompt(*, prompt: str, system_prompt: str) -> str:
     return f"{system_prompt}\n\n## Task\n{prompt}"
 
 
+def _should_inline_system_prompt(system_prompt: str) -> bool:
+    return os.name == "nt" and len(system_prompt) > _WINDOWS_ARGV_SYSTEM_PROMPT_LIMIT
```
wenchanghan: overall, do we not want to pull the os.name == "nt" out? I see this is getting checked multiple times and it's error prone

wenchanghan:
### Code review

Found 1 issue:

1. `_epoch_to_iso`'s new millisecond-detection floor exactly equals the documented "no upper limit" sentinel, silently corrupting it (bug: `_MIN_PLAUSIBLE_MILLISECOND_EPOCH_TS = 1_000_000_000_000` is exactly `10**12`, the value the adjacent comment says callers pass as an open upper bound)

https://github.com/ReflexioAI/reflexio/blob/829b67b17911be2f7a1cb210ce007fbc72c2708f/reflexio/server/services/storage/sqlite_storage/_base.py#L389-L410

`_epoch_to_iso(10**12)` now returns `2001-09-09T01:46:40+00:00` instead of the pre-PR `9999-12-31T23:59:59+00:00`, because `10**12` falls inside the new inclusive range `[_MIN_PLAUSIBLE_MILLISECOND_EPOCH_TS, _MAX_SAFE_EPOCH_TS * 1000]` and gets divided by 1000. This sentinel is live — `tests/server/api_endpoints/test_evaluation_overview_api.py#L15` passes `to_ts=1_000_000_000_000` — and is the exact value #172 clamped after it 500'd the evaluation-overview endpoint. This PR reintroduces that failure mode, just silently instead of as a crash: an "open" upper-bound query now resolves to a 2001 cutoff and would exclude nearly all real rows. The PR's new tests cover `10**18` but not `10**12`, so this isn't caught.

PR #290, opened 2026-07-04
fix: count manual profiles for all users

wenchanghan:
## Summary
- Fix manual profile generation counts when the request omits `user_id` and processes all interaction users.
- Preserve `processed_user_ids=[]` as an empty successful-user batch so failed or cancelled users do not recount stale profiles.
- Count generated profiles through a storage-layer batch count instead of hydrating profiles once per processed user.

## Broken Before
- All-user manual generation (`user_id=None`) could report generated-profile totals from the wrong scope because counting fell back to request-level user context instead of the concrete users processed by the batch.
- A zero-success batch for a specific `user_id` could still count stale pre-existing CURRENT profiles for that user, producing self-contradictory results like generated profiles for `0` processed users.
- The first fix iteration avoided the all-user overcount but introduced per-user profile hydration during counting, increasing query count and transient DB failure surface with batch size.

## Fixed Now
- Batch counting is scoped to `processed_user_ids`, with `[]` treated as an intentional empty result rather than a missing value.
- Manual generation counts CURRENT profiles and rerun generation counts PENDING profiles through the same generated-count hook used by `_run_batch_with_progress()`.
- SQLite now has a single-query `count_user_profiles_by_status(user_ids, status)` override; storage backends without an override use the backward-compatible BaseStorage fallback.

## Expected Behavior
- Manual generation for all users reports profiles generated only for users that successfully processed in that run.
- If every user fails or the operation is cancelled before any user succeeds, the generated-profile count is `0` and no stale profiles are reported as new output.
- Rerun counts continue to count PENDING profiles, but empty processed-user batches stay empty.
- Counting does not require one `get_user_profile()` call per processed user on SQLite.

## Changes
- `ProfileGenerationService._get_generated_count()` dispatches manual runs to CURRENT-profile counting and reruns to PENDING-profile counting.
- `run_manual_regular()` uses the count returned by `_run_batch_with_progress()`; the count now flows through `_get_generated_count(processed_user_ids=...)`.
- Added backward-compatible `count_user_profiles_by_status(user_ids, status)` storage support; SQLite overrides it with a single-query implementation so profile generation can count concrete users/statuses without per-user hydration.
- Empty user lists return `0` before hitting storage, preserving the no-stale-count behavior for zero-success batches.
- Removed dead test setup that implied `run_manual_regular()` exercised profile hydration directly.

## Test Plan
- `uv run pytest tests/server/services/profile/test_profile_generation_service.py tests/server/services/storage/test_storage_contract_profiles.py -q --no-cov`
- `uv run pytest tests/server/services/test_profile_generation_service.py tests/server/services/profile/test_profile_generation_service.py tests/server/services/storage/test_storage_contract_profiles.py -q --no-cov`
- `uv run ruff check reflexio/server/services/profile/service.py reflexio/server/services/storage/storage_base/profiles/_profile_store.py reflexio/server/services/storage/sqlite_storage/profiles/_profile_store.py tests/server/services/profile/test_profile_generation_service.py tests/server/services/storage/test_storage_contract_profiles.py`
- `uv run ruff format --check reflexio/server/services/profile/service.py reflexio/server/services/storage/storage_base/profiles/_profile_store.py reflexio/server/services/storage/sqlite_storage/profiles/_profile_store.py tests/server/services/profile/test_profile_generation_service.py tests/server/services/storage/test_storage_contract_profiles.py`
- `uv run pyright reflexio/server/services/profile/service.py reflexio/server/services/storage/storage_base/profiles/_profile_store.py reflexio/server/services/storage/sqlite_storage/profiles/_profile_store.py tests/server/services/profile/test_profile_generation_service.py tests/server/services/storage/test_storage_contract_profiles.py`

### DIFF
--- a/reflexio/server/services/profile/service.py
+++ b/reflexio/server/services/profile/service.py
@@ -498,14 +498,12 @@ class ProfileGenerationService(
             )
 
         user_ids = self._count_user_ids(request.user_id, processed_user_ids)
-        total = 0
-        for user_id in user_ids:
-            profiles = self.storage.get_user_profile(  # type: ignore[reportOptionalMemberAccess]
-                user_id=user_id,
-                status_filter=[Status.PENDING],
-            )
-            total += len(profiles)
-        return total
+        if not user_ids:
+            return 0
+        return self.storage.count_user_profiles_by_status(  # type: ignore[reportOptionalMemberAccess]
+            user_ids=user_ids,
+            status=Status.PENDING,
+        )
 
     @staticmethod
     def _count_user_ids(
@@ -692,6 +690,8 @@ class ProfileGenerationService(
                 "source": request.source,
                 "mode": "manual_regular",
             }
+            # total_profiles is computed inside the batch runner via
+            # _get_generated_count(processed_user_ids=...).
             users_processed, total_profiles = self._run_batch_with_progress(
                 user_ids=user_ids,
                 request=request,  # type: ignore[reportArgumentType]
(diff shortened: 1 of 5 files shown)


yyiilluu:
i thought this one is already fixed? it should be done for both oss and enterprise if needed

wenchanghan:
> i thought this one is already fixed? it should be done for both oss and enterprise if needed

current main does not have the fix. Main still has processed_user_ids or ..., _count_manual_generated(request) counting by request context, and no count_user_profiles_by_status (note that this status is about generation status).  This is only implemented for SQLite in this repo.

Other extension repos, if using diff DB, can extend the updated contract (default to have no behavior change) with separate PR

PR #292, opened 2026-07-05
fix: retry MiniLM embed after clearing corrupt cache

wenchanghan:
## Summary
- Adds Reflexio-layer self-healing for recoverable Chroma MiniLM cache failures during local embedding use.
- Serializes MiniLM cache recovery across processes so multiple backend workers, CLIs, or embedding daemons do not concurrently delete/rebuild the same on-disk cache.
- Leaves API, storage, schema, and database contracts unchanged.

## Before
- A corrupt or partial Chroma MiniLM cache could fail the first `local/minilm-l6-v2` embed call and surface as an opaque local embedding error.
- Cache recovery handled by host scripts only helped that host path; in-process local embeddings and the embedding daemon did not share the fix.
- Concurrent processes could hit the same corrupt cache and race while deleting or rebuilding `all-MiniLM-L6-v2`.
- If automatic recovery still failed, the user did not get a concrete cache-recovery next step.

## After
- Recoverable MiniLM cache failures are handled inside Reflexio's local embedding provider, so daemon warmup and in-process local embeddings use the same recovery path.
- Reflexio clears the failed embedder class's MiniLM cache and retries the embed once.
- Recovery is protected by a cross-process filesystem lock; if another process has already refreshed the cache, Reflexio retries against the refreshed cache instead of deleting it again.
- If the retry is exhausted, the raised error names the cache directory and tells the operator to delete it, restart Reflexio, and retry local embedding.
- The expected extracted-file list is documented as coupled to Chroma's private MiniLM archive layout because Chroma does not expose that list as a public runtime contract.

## Coverage
- Partial-cache retry and complete-but-corrupt-cache retry.
- Constructor/dependency failure preserving cache contents.
- Thread-local recovery reuse.
- Cross-process recovery lock coverage and already-refreshed-cache reuse.
- Retry-exhausted manual recovery message.
- Real Chroma `ONNXMiniLM_L6_V2` cache attribute shape without triggering a download.

## Related
- Supports the lower-layer Reflexio recovery needed by ReflexioAI/claude-smart#98, leaving Claude Smart to handle host-level readiness and orchestration.

## Test Plan
- `uv run pytest tests/server/llm/test_local_embedding_provider.py -q -o addopts=`
- `uv run pytest tests/server/llm/test_embedding_service_provider.py tests/server/llm/test_embedding_service_concurrency.py -q -o addopts=`
- `uv run ruff check reflexio/server/llm/providers/local_embedding_provider.py tests/server/llm/test_local_embedding_provider.py`
- `uv run ruff format --check reflexio/server/llm/providers/local_embedding_provider.py tests/server/llm/test_local_embedding_provider.py`
- `uv run pyright reflexio/server/llm/providers/local_embedding_provider.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of local MiniLM embeddings by automatically retrying once when the local embedding cache is corrupted or incomplete.
  * Added coordinated cache-recovery behavior to prevent conflicting updates and to avoid unnecessary cache deletion.
  * Enhanced failure guidance so users know how to recover when repair can’t complete.

* **Tests**
  * Expanded coverage for cache corruption/retry behavior, including lock-protected cleanup and correct error handling on retry exhaustion.

* **Documentation**
  * Updated CLI docs with step-by-step recovery instructions for MiniLM cache failures, and suggested switching providers if needed.

### DIFF
--- a/reflexio/cli/README.md
+++ b/reflexio/cli/README.md
@@ -79,8 +79,8 @@ When `REFLEXIO_EMBEDDING_PROVIDER=local_service` or
 requests allow extra time for model cold start; override with
 `REFLEXIO_EMBEDDING_SERVICE_TIMEOUT_MS` when needed.
 MiniLM cache corruption is retried once automatically. If recovery still fails,
-delete the cache directory named in the error message and restart Reflexio, or
-switch to a cloud embedding provider.
+delete the cache directory named in the error message, restart Reflexio, and
+retry local embedding.
 
 ## Publishing interactions
 
--- a/reflexio/cli/README.md
+++ b/reflexio/cli/README.md
@@ -78,6 +78,9 @@ When `REFLEXIO_EMBEDDING_PROVIDER=local_service` or
 `local/nomic-embed-text-v1.5`, and `local/minilm-l6-v2`. Local embedding
 requests allow extra time for model cold start; override with
 `REFLEXIO_EMBEDDING_SERVICE_TIMEOUT_MS` when needed.
+MiniLM cache corruption is retried once automatically. If recovery still fails,
+delete the cache directory named in the error message and restart Reflexio, or
+switch to a cloud embedding provider.
 
 ## Publishing interactions
 
--- a/reflexio/server/llm/providers/local_embedding_provider.py
+++ b/reflexio/server/llm/providers/local_embedding_provider.py
@@ -126,7 +126,7 @@ class LocalEmbedder:
         safe_inputs = [(text or "")[:_MAX_CHARS] for text in texts]
         try:
             raw = ef(safe_inputs)
-        except Exception as exc:
+        except Exception as exc:  # noqa: BLE001 - Chroma raises varied cache errors.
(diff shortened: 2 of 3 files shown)


reflexio/cli/README.md
```diff
@@ -78,6 +78,9 @@ When `REFLEXIO_EMBEDDING_PROVIDER=local_service` or
 `local/nomic-embed-text-v1.5`, and `local/minilm-l6-v2`. Local embedding
 requests allow extra time for model cold start; override with
 `REFLEXIO_EMBEDDING_SERVICE_TIMEOUT_MS` when needed.
+MiniLM cache corruption is retried once automatically. If recovery still fails,
+delete the cache directory named in the error message and restart Reflexio, or
+switch to a cloud embedding provider.
```
yyiilluu: should never switch to claude embedding provider

yyiilluu (approved):

reflexio/cli/README.md
wenchanghan: Good point thanks. 3 files needed change and all resolved.

PR #298, opened 2026-07-06
Add --skip-if-running for idempotent service starts

wenchanghan:
## What Was Annoying

Callers that legitimately want an idempotent ensure-running command had to duplicate Reflexio health checks before calling `reflexio services start`, or call it anyway and tolerate port conflicts when the selected backend or embedding service was already healthy.

## What Changed

- Add `--skip-if-running` to `reflexio services start` and the lower-level service runner.
- When the flag is used, Reflexio performs a final in-process health reprobe before deciding to skip startup.
- Limit health-based skipping to backend and embedding because they have dedicated health endpoints.
- Leave docs/frontend on normal startup and port-conflict behavior until they have reliable service health endpoints.
- Keep the flag as ensure-running behavior only; it does not reload, restart, or detect changed service code.

## Behavior After Fix

Healthy selected backend/embedding services are treated as already satisfied and are not relaunched. Unhealthy, missing, or no-longer-healthy selected services continue through the normal startup path.

If service code changes and a fresh process is required, the caller must make that restart decision explicitly, such as with `reflexio services stop` followed by `reflexio services start`, or with caller-owned upgrade/restart logic before invoking this ensure-running path.

## Positioning

This is a good-to-have generic CLI idempotency improvement for Reflexio. It is not a required dependency for claude-smart update/reload behavior, and it does not move host-specific identity, version, or compatibility decisions into Reflexio.

## Validation

- `uv run pytest --no-cov tests/cli/test_utils.py tests/cli/test_run_services_workers.py tests/cli/test_services_first_run.py -q` -> 36 passed
- `uv run ruff check reflexio/cli/commands/services.py reflexio/cli/run_services.py reflexio/cli/utils.py tests/cli/test_utils.py tests/cli/test_run_services_workers.py tests/cli/test_services_first_run.py` -> passed
- `uv run pyright reflexio/cli/commands/services.py reflexio/cli/run_services.py reflexio/cli/utils.py tests/cli/test_utils.py tests/cli/test_run_services_workers.py tests/cli/test_services_first_run.py` -> 0 errors
- `git diff --check` -> passed

### DIFF
--- a/reflexio/cli/commands/services.py
+++ b/reflexio/cli/commands/services.py
@@ -20,6 +20,26 @@ _logger = logging.getLogger(__name__)
 app = typer.Typer(help="Start and stop Reflexio services.")
 
 
+def _backend_already_healthy_for_skip(
+    *,
+    backend_port: int | None,
+    docs_port: int | None,
+    embedding_port: int | None,
+) -> bool:
+    """Return True when ``--skip-if-running`` can bypass backend setup prompts."""
+    from reflexio.cli.utils import service_port_healthy
+
+    ports = run_mod.resolve_ports(
+        argparse.Namespace(
+            backend_port=backend_port,
+            docs_port=docs_port,
+            embedding_port=embedding_port,
+        ),
+        defaults=run_mod.DEFAULT_OSS_PORTS,
+    )
+    return service_port_healthy("backend", ports["backend"])
+
+
 def _ensure_llm_configured(env_path: Path) -> None:
     """Run the first-run LLM + embedding wizard when no key is configured.
 
@@ -218,6 +238,16 @@ def start(
             help="Seconds to drain in-flight requests on shutdown.",
         ),
     ] = 30,
+    skip_if_running: Annotated[
+        bool,
+        typer.Option(
+            "--skip-if-running",
+            help=(
+                "Skip selected services whose health endpoint is already responding. "
+                "Does not detect code changes; stop then start to reload."
+            ),
+        ),
+    ] = False,
 ) -> None:
     """Start Reflexio services (backend, docs)."""
     from reflexio.cli.bootstrap_config import resolve_storage, save_storage_to_config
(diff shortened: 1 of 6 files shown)


yyiilluu:
Suppose the service code got changed. Does that mean that you will not pick up the updated change?
How to pick up updates now?

wenchanghan:
> Suppose the service code got changed. Does that mean that you will not pick up the updated change? How to pick up updates now?

Clarifying the customer-facing contract here: --skip-if-running is intentionally an idempotent ensure-running flag, not a reload flag. In generic Reflexio, if the selected service is already healthy, this flag leaves it alone; callers that need new code must use explicit restart semantics such as reflexio services stop --only backend && reflexio services start --only backend.

For host-managed integrations like claude-smart, the better UX is not to make users know that Reflexio detail. claude-smart has runtime identity/version headers, so its service manager can detect “healthy claude-smart service, but old plugin/service version” and restart that owned local service during plugin startup. I’m handling that in the claude-smart PR layer; Reflexio stays conservative and predictable.

yyiilluu:
not getting this, if the claude-smart always start reflexio using `--skip-if-running` flag, then how does the claude-smart update flow work now

wenchanghan:
The update flow is handled one layer above this flag, in claude-smart's service manager.

Sequence in the claude-smart PR:

1. `backend-service.sh start` acquires the claude-smart service lock and probes the existing backend's marker headers.
2. If the running backend is marked as claude-smart but reports an older or missing `x-claude-smart-version`, claude-smart stops that owned local backend/embedding service first.
3. Only after that decision does it invoke `reflexio services start --skip-if-running --only backend ...`.

So `--skip-if-running` is not the update mechanism. It is the final idempotent launcher guard: if a compatible backend is already healthy, Reflexio exits cleanly; if claude-smart already stopped an older backend, there is nothing healthy to skip and Reflexio starts the new service code.

For plain Reflexio CLI usage, the contract stays simple: `--skip-if-running` means ensure-running, not reload. For claude-smart, plugin update/restart semantics live in claude-smart because it has the plugin-version marker needed to make that decision safely.

yyiilluu:
Isn't this kind of weird? If claude-smart (which is the caller) knows that Reflexio is running correctly, why is it calling start and skip if running? You should just check on claude-smart's side.
also, i have another change coming that might reverted some changed i made previously in https://github.com/ReflexioAI/claude-smart/pull/120

wenchanghan:
### Code review

Found 1 issue:

1. String built with `+` concatenation instead of an f-string (`.claude/rules/pythonic-code.md` says "Always use f-strings for general string interpolation")

https://github.com/ReflexioAI/reflexio/blob/630eb5afb2430966dab0dc62d14054712c6625d1/reflexio/cli/utils.py#L363-L367

wenchanghan:
After reviewing the caller/update semantics across Reflexio and claude-smart, we decided not to merge this PR.

The implementation is workable, but the feature pushes Reflexio's generic service start command into edge-case handling that is better avoided right now. Keeping Reflexio's CLI behavior simple and leaving caller-specific start/update decisions in the caller layer keeps both repositories leaner and easier to reason about.

Thanks for the discussion and review. Closing this intentionally rather than carrying the extra complexity forward.

PR #326, opened 2026-07-10
feat(scheduling): Workstream B Phase 1 — leader-gate seam, bounded org fan-out, shared callback executor

guangyu-reflexio:
## Summary

Phase 1 (OSS half) of the Workstream B scheduler-coordination design (spec: `docs/superpowers/specs/2026-07-10-workstream-b-scheduler-coordination-design.md` in the enterprise repo; review-converged over 3 rounds). Prepares the daemon fleet for multi-instance deployments and 1,000-org scale:

1. **`LeaderGate` seam on `ThreadedScheduler`** (`scheduling.py`, stays stdlib-only): an optional injected fleet-coordination gate consulted per tick. `None` (OSS default) preserves today's always-tick behavior byte-for-byte; a follower waits a fixed 60s poll; a gate that violates its never-raise contract is caught and **fails open** (tick runs — duplicate work, never silence). The enterprise layer supplies the real advisory-lock gate in the follow-up PR via the new `set_leader_gate` module hook (mirrors `set_org_id_provider`).
2. **`iterate_orgs_bounded`** (`org_fanout.py`, new): bounded per-tick parallel org iteration with per-org timeout, adopted by `LineageGCScheduler` (`REFLEXIO_SCHEDULER_ORG_WORKERS`, default 8; SQLite pinned serial). Timeout classification is honest: only orgs that *started and exceeded* their run budget are `timed_out`; orgs starved behind stuck workers log `org_sweep_starved` and never poison the repeat-timeout anomaly (`lineage.gc.org_sweep_timeout_repeat`, fired only on strictly-consecutive repeats).
3. **`BoundedCallbackExecutor`** (`callback_executor.py`, new): one process-wide 16-worker pool with a 256-entry drop-oldest queue replacing thread-per-fire in the three debounce schedulers (tagging, group evaluation, playbook optimization). Thread count goes O(active keys) → O(1). Drop logging happens outside the lock; sustained overflow (>10 drops/min) escalates a throttled anomaly.
4. **Multi-worker visibility**: one WARNING at data-plane startup when the server runs >1 worker process (daemons tick in every worker; safe by the concurrent-tick invariant).

## Intended behavior change (reviewer note)

The three converted schedulers previously spawned an unbounded thread per fire. They now share the bounded executor: under a sustained 256-deep backlog, **drop-oldest evicts the oldest queued fire**, and a dropped fire for a key with no subsequent event is skipped. This is a design-accepted best-effort tradeoff (spec §6.2 — owned for all three families), swapping a silent-OOM failure mode for a bounded, alarmed one.

## Testing

- 3,568 unit tests pass; the 7 failures in `tests/server/services/governance/test_subject_write_barrier_sqlite.py` are **pre-existing and red on `origin/main`** (verified against the unmodified base twice; zero overlap with this diff).
- e2e tier: 43 passed. Ruff + Pyright clean on all changed files.
- New coverage: gate matrix (None/True/False/raise), fan-out concurrency bound (upper+lower), stuck/starved-org classification, pooled stop-event path, env-parsing fallbacks, alternating-tick non-escalation, module-hook fallback, executor drop-oldest/drop-rate/exception-survival, playbook-optimizer real fire path. Threading-heavy files re-run 5x, no flakes.
- Reviewed by a 5-lens review loop (correctness, security-resilience, architecture, verification-testing, docs-api): 18 findings fixed and independently verified in one iteration.

## Enterprise follow-up (do not merge out of order)

This PR is self-contained and inert without the enterprise half: no gate is injected (all daemons keep today's behavior except the LineageGC fan-out, the executor conversion, and the startup log). The enterprise PR (coordination module, leadership table, wiring, submodule repin) lands next per the design's §10.1 order.

## Summary by CodeRabbit

* **New Features**
  * Added configurable bounded parallel processing for scheduled organization maintenance tasks.
  * Added leader coordination to prevent duplicate scheduler activity across worker processes.
  * Added resilient background callback execution with bounded queuing and overflow handling.
  * Added startup warnings for multi-worker daemon configurations.
  * Added optional scheduler worker-count configuration, with serial execution preserved for SQLite.

* **Bug Fixes**
  * Improved timeout classification and cancellation handling for maintenance tasks.
  * Prevented callback failures from stopping subsequent scheduled work.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -86,6 +86,10 @@ REFLEXIO_MISSING_VECTOR_BACKFILL_ENABLED=false
 # Max interactions re-embedded per org per sweep tick (bounds embedding cost and
 # log volume). (optional; default 200; values <= 0 fall back to the default)
 REFLEXIO_MISSING_VECTOR_BACKFILL_CAP=200
+# Bounded per-tick org fan-out width for the lineage-GC / expiry-reclamation
+# scheduler. SQLite storage is pinned to serial (1) regardless of this value.
+# (optional; default 8; non-numeric or <=0 falls back to default)
+REFLEXIO_SCHEDULER_ORG_WORKERS=
 
 # =============================================================================
 # 5. TESTING & LOGGING
--- a/reflexio/server/api.py
+++ b/reflexio/server/api.py
@@ -16,7 +16,6 @@ iterates (e.g. the QPS billable-endpoint scan over ``core_router.routes``).
 
 import inspect
 import logging
-import os
 from collections.abc import Callable, Iterable
 from typing import TYPE_CHECKING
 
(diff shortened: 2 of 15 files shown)


guangyu-reflexio:
Thanks — triaged the nitpick on `test_scheduler_fires_callback_through_executor`. Leaving it as-is, with rationale:

- `PlaybookOptimizationScheduler` (like its siblings `TaggingScheduler`/`GroupEvaluationScheduler`) is a **process-lifetime singleton with no `stop()` API** by design — the suggested `scheduler.stop()` doesn't exist on it. Adding a shutdown path to production code solely for this test would be scope creep against the design (spec §6.2 treats these as process-wide).
- The accepted sibling test `tagging/test_tagging_scheduler.py::test_scheduler_fires_scheduled_callback` leaks the same way (via the singleton). This test deliberately constructs a *fresh* instance to avoid polluting the class singleton, at the cost of one extra daemon thread.
- After the callback fires, the queue is empty, so the loop **parks on `_wake_event.wait()` (blocked, ~zero CPU)** rather than spinning — it does not "continue ticking."

Net: a single parked daemon thread, consistent with existing convention, not worth a production API addition. Happy to revisit if a `stop()` is added to this scheduler family in a later phase.

PR #331, opened 2026-07-11
docs(developer): document isolation rule for future self-bootstrapping test harnesses

wenchanghan:
### Why

`claude-smart` recently added isolation guards (PR #125) to its self-bootstrapping integration harness because that harness boots services on the production ports (`8071`/`8072`/`3001`) from a clean checkout — risking clobbering a developer's installed instance.

Reflexio OSS has the same potential hazard: `reflexio services start` defaults to ports `8061`/`8062`, and a future CI harness or contributor script could bind those ports and replace a developer's local production instance.

Today the E2E suite avoids the problem by running `Reflexio` in-process with a `tmp_path` SQLite fixture (see `tests/e2e_tests/conftest.py`), so no port binding and no `~/.reflexio` writes occur.

### What

Document the rule for any *future* self-bootstrapping harness: sandbox storage, use non-default ports, refuse real `$HOME`. This is documentation only — no behavior changes.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for building self-bootstrapping end-to-end test harnesses.
  * Documented safe sandboxing requirements, including isolated storage, non-default ports, and avoiding the real home directory.
  * Included context for preventing test harnesses from interfering with production services.

### DIFF
--- a/developer.md
+++ b/developer.md
@@ -179,10 +179,10 @@ Today the E2E suite runs `Reflexio` in-process with `StorageConfigSQLite` config
 If you add a future harness that boots services from a clean checkout — analogous to claude-smart's `tests/integration/integration.sh` — it must:
 
 1. Sandbox storage: point `LOCAL_STORAGE_PATH` (or the SQLite `db_path`) at a temp dir, never the default `~/.reflexio/data/`.
-2. Use non-default ports: `BACKEND_PORT=19061`, `DOCS_PORT=19062` (or higher), and refuse production ports (`8061`, `8062`) unless the user explicitly opts in.
-3. Refuse to use the real `$HOME` as the integration home; create a temp `INTEG_HOME` and export it before launching services.
+2. Use non-default ports: pick a `1XXXX` form that mirrors the production `8061`/`8062` while staying clear of common dev ranges (e.g. `BACKEND_PORT=19061`, `DOCS_PORT=19062`, or higher), and refuse production ports (`8061`, `8062`) unless the user explicitly opts in.
+3. Never use the real `$HOME` as the integration home; create a temp `INTEG_HOME` and export `HOME=$INTEG_HOME` before launching services so every subprocess inherits the isolated home.
 
-Rationale: claude-smart learned this the hard way when its harness bound production ports (`8071`/`8072`/`3001`) and either failed or replaced the user's installed instance — see [claude-smart PR #125](https://github.com/ReflexioAI/claude-smart/pull/125) (`test: isolate integration harness runtime`). OSS avoids the hazard today only because no such harness exists yet; the above rule prevents reintroducing it when one is added.
+Rationale: claude-smart encountered this when its harness bound production ports (`8071`/`8072`/`3001`) and either failed or replaced the user's installed instance. OSS avoids the hazard today only because no such harness exists yet; the above rule prevents reintroducing it when one is added.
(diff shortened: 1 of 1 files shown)


wenchanghan:
**Suggestion (minor):** The link to `ReflexioAI/claude-smart#125` is accessible, but a specific PR reference feels too granular for developer docs — the rationale prose already carries the point. Consider dropping the hyperlink and keeping just the text, e.g.:

> …claude-smart encountered this when its harness bound production ports (`8071`/`8072`/`3001`) and either failed or replaced the user's installed instance.

That keeps the historical context without anchoring the docs to a specific PR that future readers will have to chase.

---

**Suggestion (nit):** `BACKEND_PORT=19061` / `DOCS_PORT=19062` — the choice of those numbers isn't obvious. A brief parenthetical like "mirrors the production ports with a leading `1`" would make the rationale legible to a contributor picking ports for a new harness.

---

**Nit:** "refuse to use the real `\$HOME`" reads oddly — phrased as if the harness itself refuses. "never use the real `\$HOME` as the integration home" matches the imperative pattern of the other two items.

---

Everything else looks good — the factual claims (in-process `StorageConfigSQLite` + `tmp_path`, no port binding, no `~/.reflexio` writes) all check out against `tests/e2e_tests/conftest.py`.

wenchanghan:
Addressed all three suggestions plus the CodeRabbit inline note in commit `202590d`:

1. **Drop the PR hyperlink** — `developer.md` had no other `[text](url)` patterns; the rationale prose carries the point. Removed both the URL and the trailing commit-message parenthetical.
2. **Port rationale** — clarified that `1XXXX` mirrors `8061`/`8062` while staying clear of common dev ranges (`3000`, `5000`, `8000`, `8091`, etc.).
3. **Wording** — rule (3) now reads "Never use the real `$HOME` as the integration home…"
4. **CodeRabbit fix (executed)** — exported `HOME=$INTEG_HOME` instead of just `INTEG_HOME`, so every subprocess inherits the isolated home and none can fall back to the real `~/.reflexio` via `$HOME` resolution.

Diff (3 lines changed):

```diff
-2. Use non-default ports: `BACKEND_PORT=19061`, `DOCS_PORT=19062` (or higher), and refuse production ports (`8061`, `8062`) unless the user explicitly opts in.
-3. Refuse to use the real `$HOME` as the integration home; create a temp `INTEG_HOME` and export it before launching services.
+2. Use non-default ports: pick a `1XXXX` form that mirrors the production `8061`/`8062` while staying clear of common dev ranges (e.g. `BACKEND_PORT=19061`, `DOCS_PORT=19062`, or higher), and refuse production ports (`8061`, `8062`) unless the user explicitly opts in.
+3. Never use the real `$HOME` as the integration home; create a temp `INTEG_HOME` and export `HOME=$INTEG_HOME` before launching services so every subprocess inherits the isolated home.

-Rationale: claude-smart learned this the hard way when its harness bound production ports (`8071`/`8072`/`3001`) and either failed or replaced the user's installed instance — see [claude-smart PR #125](https://github.com/ReflexioAI/claude-smart/pull/125) (`test: isolate integration harness runtime`). …
+Rationale: claude-smart encountered this when its harness bound production ports (`8071`/`8072`/`3001`) and either failed or replaced the user's installed instance. …
```

Resolving comment thread.

PR #332, opened 2026-07-11
fix(llm): drop local/* in-process embedding models from generation fallbacks

guangyu-reflexio:
## Summary

Filters `local/*` in-process embedding models out of text-generation fallback lists so they can never be handed to litellm's completion fallback ladder.

## Problem

`local/*` models (e.g. `local/nomic-embed-text-v1.5`) are **in-process ONNX embedding** models — `_litellm_embedding.py` routes them to `NomicEmbedder`/`LocalEmbedder` and they have **no litellm completion route**. If such a model ends up in a text-generation `fallback_models` list, `_build_completion_params` passes it straight into litellm's `fallbacks`, and when the primary model fails, litellm tries the embedding model via `async_completion_with_fallbacks` → `get_llm_provider` → **`BadRequestError: LLM Provider NOT provided. You passed model=local/nomic-embed-text-v1.5`**.

This surfaced in production as **Sentry PYTHON-FASTAPI-CV (~1,290 events)**. The existing filter at `_build_completion_params` only deduped the primary model (`m != actual_model`); there was no guard against non-generation model names.

## Fix

`_build_completion_params` now drops `local/*` entries alongside the existing primary-model dedup:

```python
fallback_models = [
    m for m in fallback_models_raw
    if m != actual_model and not m.startswith("local/")
]
```

`local/` is the sole in-process embedding prefix (`embedding_service.py`), and no generation model uses it, so this is safe and targeted — a stray embedding model in the fallback list is now skipped instead of crashing the call.

## Test

Added `test_local_embedding_model_is_dropped_from_fallbacks`: a `local/nomic-embed-text-v1.5` + `gpt-5-mini` fallback list resolves to `["gpt-5-mini"]`. All existing fallback tests still pass (10 passed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text-generation fallback handling by excluding `local/*` embedding models from the fallback ladder, avoiding provider routing errors when mixed with valid generation models.

* **Tests**
  * Added unit coverage to confirm `local/*` models are filtered out while valid fallbacks are preserved.
  * Added a regression/contract test to ensure the `local` provider is embedding-only and excluded from generation-capable provider defaults.

### DIFF
--- a/reflexio/server/llm/_litellm_text_generation.py
+++ b/reflexio/server/llm/_litellm_text_generation.py
@@ -325,8 +325,16 @@ class TextGenerationMixin:
         }
 
         # Drop any fallback entry that points back at the primary — sending the
-        # same broken endpoint twice never helps.
-        fallback_models = [m for m in fallback_models_raw if m != actual_model]
+        # same broken endpoint twice never helps. Also drop in-process ``local/*``
+        # embedding models: they have no litellm completion route (they are served
+        # in-process by ``_litellm_embedding.py``), so litellm would raise
+        # ``BadRequestError: LLM Provider NOT provided`` deep inside its fallback
+        # ladder if one ever landed in this list (Sentry PYTHON-FASTAPI-CV).
+        fallback_models = [
+            m
+            for m in fallback_models_raw
+            if m != actual_model and not m.startswith("local/")
+        ]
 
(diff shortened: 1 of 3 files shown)


guangyu-reflexio:
@coderabbitai review

PR #333, opened 2026-07-11
feat(storage): preserve retention evidence for later evaluation

wenchanghan:
## Summary

Reflexio bounds live SQLite tables by deleting old rows. This PR optionally preserves those rows in a bounded JSONL FIFO so older sessions remain available for later evaluation without allowing either storage path to grow forever.

- Live SQLite retention remains the priority and never waits on archive recovery.
- New evidence is written first; oldest complete archive segments are then evicted until total JSONL size is under the configured ceiling.
- Each segment keeps one retention batch together, including cascade-deleted rows.
- Interaction rows retain every current evaluation input; only embeddings are excluded.
- The publish schema accepts bounded injection-time learning snapshots so later evaluation does not substitute edited learning text.

## Flow

```mermaid
flowchart LR
    A["SQLite selects up to 1,000 old target rows"] --> B{"Does the batch fit?"}
    B -->|Yes| C["Write one JSONL segment"] --> D["Delete the same live rows"]
    B -->|No| E["Split into smaller complete batches"] --> C
    C --> F["If over the ceiling, evict oldest segments"]
    F --> G["Next PR: read live + JSONL into evaluator"]
```

## Archive contract

| Setting | Behavior |
|---|---|
| `REFLEXIO_RETENTION_ARCHIVE=1` | Enable archive-before-delete for SQLite. Default is off. |
| `REFLEXIO_RETENTION_ARCHIVE_MAX_BYTES=10737418240` | Set total completed-segment ceiling. Default is 10 GiB. |

Archive files always live in an `archive` directory beside the SQLite database—for example, `~/.reflexio/data/archive` beside the default database.

Each archive-enabled SQLite transaction selects at most 1,000 retention-target rows. This is a work-size bound, not a failure threshold. A timestamped `*.jsonl` segment contains those rows and any related rows deleted with them. For example, deleting an old `requests` row also deletes and archives the `interactions` rows that reference its `request_id`; many targets have no related rows and remain single-table batches.

If a batch expands beyond the byte ceiling, Reflexio recursively splits it into smaller complete segments instead of dropping the entire batch. FIFO then retains the newest complete groups that fit. Only one indivisible target plus its related rows can be skipped if that group alone exceeds the entire archive ceiling. Archive I/O failures are also logged. In all cases, live retention continues so learning and application behavior remain available and SQLite stays bounded.

## Example records

One segment may contain several tables; every line identifies its table explicitly.

```jsonl
{"table":"requests","archived_at":1783792800,"row":{"request_id":"req-123","user_id":"project-7","session_id":"session-42","created_at":1783789200,"source":"agent","agent_version":"v1"}}
{"table":"interactions","archived_at":1783792800,"row":{"interaction_id":987,"request_id":"req-123","role":"Assistant","content":"Use pathlib here.","retrieved_learnings":"[{\"kind\":\"user_playbook\",\"learning_id\":\"11\",\"snapshot\":{\"title\":\"Filesystem discipline\",\"content\":\"Use pathlib for filesystem work.\",\"trigger\":\"writing Python filesystem code\",\"rationale\":\"Path objects are easier to compose.\"}}]"}}
```

### Why the complete interaction row is retained

This is evaluation evidence, not an additional application log. Existing evaluation paths use more than message text:

- agent-success evaluation uses content, roles, tool calls, user actions, and the final image;
- shadow evaluation uses `shadow_content`;
- learning attribution uses citations and `retrieved_learnings`.

Dropping those fields here would make some retroactive evaluations impossible. The large vector `embedding` column is excluded because no evaluator needs it. Tests enforce “all interaction columns except embedding.”

The learning snapshot contains the injection-time source fields associated with that response. It lets a later evaluator avoid substituting a learning that has since been edited.

## What comes next

The evaluator follow-up only needs to:

1. Read request and interaction records from live SQLite and all JSONL segments.
2. Merge and deduplicate them by stable database keys.
3. Reconstruct the bounded session and use captured learning snapshots when present.
4. Run the existing evaluators and persist results normally.

## Verification

- Full suite: **4,757 passed**, 124 skipped; **82.65% coverage**.
- Focused retention/FIFO suite: **19 passed**.
- Ruff and Pyright pass on changed Python files.
- FIFO test proves the oldest segment is removed, the newest remains, and total size stays within the ceiling.
- Cascade test proves one segment keeps request and dependent interaction rows together.
- A 1,200-row integration case leaves 960 rows live, archives 240, and reconstructs all 1,200 IDs.

## Boundaries

- FIFO intentionally provides a recent bounded evidence window, not permanent history.
- JSONL is a separate export: deleting a user from SQLite does not rewrite already archived lines.
- A partial operating-system failure can leave duplicate evidence; the reader must deduplicate by table key.
- Archive export is implemented for SQLite. Other storage backends continue their existing retention behavior if archive export is unavailable.


docs/adr/0001-retention-archives-rows-before-deleting.md
```diff
@@ -0,0 +1,35 @@
+# Row-count retention archives doomed rows to JSONL before deleting
```
wenchanghan: i don't think we need to add this document.
If we explain these concisely in PR description, this should be easy to understand

wenchanghan: in fact, there shouldn't be references to claude-smart (one of the clients for Reflexio OSS) throughout the repo

wenchanghan:
**Review findings** (critical pass; the archive suite is green locally — 11/11 in `test_retention_archive.py`).

**1. Consider splitting the snapshot feature into its own PR.**
`b23091d` (`RetrievedLearningSnapshot`) is a wire-schema/evaluation change riding a storage-retention PR, and it reverses the field's documented design intent ("deliberately minimal — just the identity pair… callers should not need to supply injection-time debug fields"). It is already an isolated commit; splitting keeps this PR single-purpose and lets the schema change be weighed on its own merits (it denormalizes learning content into every interaction row — the growth this retention work exists to manage).

**2. Accepted payload bytes are unbounded.**
`snapshot.content` allows 1 MB, and `validate_retrieved_learnings_total` caps count (1000/publish) but not bytes → roughly 1 GB accepted per publish request, all persisted into the interaction row. Real learning contents are a few KB; suggest a much tighter content cap (e.g. 100 KB) or an aggregate-bytes validator.

**3. Flag-on cleanup holds the global write lock across file I/O — costly at default caps.**
With `REFLEXIO_RETENTION_ARCHIVE=1`, select → fetch → JSONL append → delete all run inside `BEGIN IMMEDIATE` plus the storage lock. At the default 250 000-row cap a single pass dooms ~50 000 rows, meaning hundreds of MB written under a global write fence — and `_retention_fetch_rows` does `SELECT *`, dragging sizable embedding blobs out of SQLite only for `append_archive_rows` to discard them. Fine for low-cap deployments; at defaults it is a multi-second stall for every writer. Suggest documenting the tradeoff in the PR body at minimum; cheap wins: select explicit columns (skip `embedding`) and/or bound rows archived per pass.

**4. Simplification opportunities.**
- The `REFLEXIO_RETENTION_ARCHIVE_WARN_BYTES` machinery — about half of `retention_archive.py`, a process-global `_warned_archive_dirs` set, and three tests — exists to emit one log line, and emits it at `logger.error` for what is semantically a warning threshold. Logging the archive size at info on each append gets most of the value for ~5 lines.
- Triple-layer chunking: the mixin chunks keys (`chunked(parent_ids)`, `chunked(keys, rows_per_batch)`) and `_retention_fetch_rows` chunks again internally (`_select_in_chunks` / its own `rows_per_chunk`). Letting the hook own chunking removes the outer loops and the `rows_per_batch` math from the mixin.

**5. Gaps.**
- The three new env vars are undocumented.
- The composite-key fetch path (e.g. a `retrieved_learning_evaluation` cleanup) and the non-SQLite fail-closed behavior (mixin defaults raising `NotImplementedError`) are untested.
- `parent_ids = [(key[0],) for key in keys]` assumes single-key cascade parents — true today and documented on `CascadeRef`, but an assert would make a future violation loud instead of a silent mis-archive.

Also verified good in this pass: fail-closed holds end-to-end (fetch errors re-raise as `StorageError`, the guard rolls back, rows survive archive failure); flag-off behavior is byte-identical, with a test asserting the guard is never acquired; all real-data cascade dependents are covered by `RETENTION_CASCADES` (fts/vec rows are derived data); union-invariant, cascade, concurrency, and dir-override tests all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

docs/adr/0001-retention-archives-rows-before-deleting.md
wenchanghan: Removed the ADR and all client-specific references from the Reflexio repository; the concise behavior and tradeoffs now live in the PR description.

wenchanghan:
Incorporated the actionable review items in `f02f21d`:

- capped snapshot content at 100 KB and aggregate snapshot text at 10 MiB per publish;
- preserved the original retention amount while splitting archive work into 1,000-row SQLite transactions;
- stopped selecting/loading embeddings during archival;
- removed redundant outer fetch chunking and added a loud assertion for cascade-key assumptions;
- documented all archive environment variables and the SQLite lock/batching tradeoff;
- added composite-key, unsupported-backend, aggregate-size, and bounded-batch coverage.

I kept the warning at error severity because the agreed product behavior is a loud operational signal once the unbounded JSONL archive crosses its threshold. I also kept the snapshot commit in this connected PR set: capturing injection-time content is required now so the follow-up can focus solely on reading JSONL into the evaluator. Full suite: 4,751 passed.

wenchanghan:
**Re-review** of the follow-up commits (`f02f21d..8d2196b`).

Prior findings all addressed:

- **Payload bounds** — snapshot content cap 1 MB → 100 KB plus a 10 MiB aggregate snapshot-bytes validator per publish, with tests.
- **Lock hold** — archive-enabled cleanup now loops in 1000-parent batches (test proves 50 × 1000 calls for a 50 k delete), and `_retention_fetch_rows` projects columns via `PRAGMA table_info` so `embedding` never leaves SQLite (test asserts the exact column set).
- **Simplifications/gaps** — chunking flattened into the hook, single-key cascade assert added, composite-key fetch and non-SQLite paths now tested, README documents the three env vars.

The revision also changes two policies rather than just fixing findings; both are implemented and documented consistently, but they invert the earlier contract, so flagging them explicitly for reviewers:

1. **Fail-open.** Archive failure now logs "evidence was not preserved" and deletion proceeds (the previous test asserting rows survive a failed archive was deliberately inverted). This removes the wedge risk — an unwritable dir or full disk can no longer stall retention — at the cost of a new worst case: a permanently misconfigured archive dir silently destroys evidence on every cleanup with only server-log visibility. Non-blocking suggestion: surface archive write failures somewhere more visible than logs, since this is now the primary evidence-loss mode.
2. **Bounded FIFO.** The archive is now per-batch segment files with oldest-whole-segment eviction at `REFLEXIO_RETENTION_ARCHIVE_MAX_BYTES`. Mechanics look sound: atomic tmp-then-rename install, the newest segment is never evicted, one segment = one retention batch so parent and cascade rows never split, zero-padded sortable names, eviction covered by tests. Consumers should note the invariant is now "most recent ~10 GiB of evictions," not "full history."

Two minor residuals, neither blocking:

- Cascade rows are unbounded per batch: 1000 parents can fan out to arbitrarily many dependent rows, all materialized in memory and encoded as a single segment under the write guard. Today's realistic fan-out (~2 interactions per request) makes this theoretical; a fan-out note or cap could come later.
- A crash between segment write and rename leaves a `.tmp` file that is never cleaned and never counted toward the ceiling (the glob is `*.jsonl`), so it lingers as junk bytes.

Verified locally at `8d2196b`: 31/31 across `test_retention_archive.py` + `test_input_size_bounds.py` (the coverage-percentage failure is the repo-wide gate applied to a partial run).

The snapshot schema remains bundled with the archive change — previously noted, maintainers' call. Otherwise ready for lead review from my side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

reflexio/server/services/storage/retention_archive.py
```diff
@@ -0,0 +1,133 @@
...
+        return False
+
+    archive_dir.mkdir(parents=True, exist_ok=True)
+    segment = archive_dir / f"{time.time_ns():020d}-{uuid4().hex}.jsonl"
+    temporary = segment.with_suffix(".tmp")
+    try:
+        temporary.write_bytes(encoded)
+        temporary.replace(segment)
+    finally:
+        temporary.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
wenchanghan: Fixed in d3ad038. Abandoned .tmp files older than one hour are swept before each append, while fresh temporary files are preserved so concurrent writers are not disrupted. Added regression coverage for both stale removal and active-file preservation.

reflexio/server/services/storage/retention_archive.py
```diff
@@ -0,0 +1,133 @@
...
+    for oldest in segments:
+        if total_bytes <= max_bytes:
+            break
+        if oldest == segment:
+            continue
+        size = oldest.stat().st_size
+        oldest.unlink()
+        total_bytes -= size
+        evicted_files += 1
+        evicted_bytes += size
```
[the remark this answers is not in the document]
wenchanghan: Fixed in d3ad038. Segment discovery and FIFO eviction now treat FileNotFoundError as a concurrent no-op while preserving all other filesystem errors. Added a regression test that removes the oldest segment immediately before unlink.

wenchanghan:
Final archive-contract clarification:

- **1,000 is a work-size cap, not a failure threshold.** Archive-enabled selection is capped at 1,000 retention-target rows per transaction.
- A batch may expand through related rows (for example, a request and its interactions). If its encoded segment exceeds the byte ceiling, Reflexio now recursively splits it into smaller complete segments instead of dropping the whole batch. Only one indivisible target plus its related rows can still be skipped if that group alone exceeds the entire ceiling.
- The archive location is intentionally fixed at `<SQLite database directory>/archive`; the unused directory override was removed.
- Segments may contain multiple tables, with every JSONL line carrying its table name.

wenchanghan:
Closing — superseded by a direction change

PR #357, opened 2026-07-17
Persist observed model provenance on learning lineage

yilu331:
## Summary

Persist the model and provider actually observed on completed LLM responses in learning-lineage events. This is an ingestion-only change: no new HTTP read/display surface and no historical backfill.

The storage invariant is now: **every durable profile, user-playbook, or agent-playbook insertion has exactly one atomic origin-lineage event**. Callers may provide per-record attribution; otherwise the canonical save synthesizes a `create` event from entity fields and leaves `model_name` / `provider` null.

Paired enterprise storage change: [ReflexioAI/reflexio-enterprise#736](https://github.com/ReflexioAI/reflexio-enterprise/pull/736)

## What changed

- Add nullable `model_name` and `provider` to lineage schemas and the SQLite migration. Existing rows remain null.
- Observe provenance only from completed response metadata. Unknown providers remain null; CLI binaries, gateways, transports, `claude`, and `codex` are not treated as providers.
- Keep existing text/chat methods returning their existing payloads. Add explicit provenance-returning entry points backed by the same typed completion implementation.
- Preserve provenance through fallback models, structured-output repair, saturation, resumable output, and failure salvage without dynamic exception attributes.
- Carry aggregation provenance structurally with each generated playbook; distinct cluster completions cannot exchange attribution.
- Construct `LineageContext` directly at generation, consolidation, and aggregation boundaries; remove the single-use provenance adapter module.
- Make `save_user_playbooks`, `add_user_profile`, and `save_agent_playbooks` the only durable save paths. Optional `lineage_contexts` carry attribution; omitted contexts still create atomic origin events with null model/provider.
- Remove `*_with_lineage` and aggregate-save wrappers. Agent aggregation now uses canonical `save_agent_playbooks` with an `aggregate` context and the shared aggregate-reason prefix.
- Make playbook edit successor creation, revise lineage, and supersede CAS one transaction. A lost CAS rolls everything back and returns `-1` without compensating deletes or ghost events.
- Decode committed outputs as legacy when no marker exists, accept version 1, and reject explicit future versions.

## Compatibility

- Existing text/chat return types are unchanged.
- Existing canonical save calls remain valid, but now always emit an origin event.
- There is intentionally no public save-without-lineage path.
- Historical lineage is neither guessed nor backfilled.

## Verification

- Ruff and `git diff --check`: clean
- Pyright: changed production/test paths clean; enterprise root check is 0 errors/warnings
- Focused lineage/LLM/storage and migration suites: passing
- Full OSS non-E2E tier: 4,901 passed, 73 skipped
- OSS E2E regressions re-run after updating stale mocks: passing
- Local service and live PostgreSQL persistence/atomicity path: passing

## Merge order

1. Merge this PR.
2. Re-pin enterprise PR #736 to the landed OSS commit.
3. Merge #736 promptly so every backend implements the same canonical-save invariant.


yilu331:
## Review — model provenance in learning lineage

Read the full diff. The lineage/storage/schema plumbing is sound and well-tested. One correctness regression and a few simplification opportunities that speak to the size of this change.

### 1. Correctness — consolidator salvage path narrowed (high)

`reflexio/server/services/playbook/components/consolidator.py` (~L749-756):

```python
except StructuredOutputRepairError as exc:
    if first_parsed_output is not None and exc.initial_provenance is not None:
        self.model_provenance = exc.initial_provenance
        return first_parsed_output
    raise
except LiteLLMClientError:
    raise
```

This changes error behavior in two ways vs. the previous `except (StructuredOutputRepairError, LiteLLMClientError): if first_parsed_output is not None: return first_parsed_output`:

- **`LiteLLMClientError` no longer salvages.** Previously, if an early turn produced a usable `first_parsed_output` and a later turn hit a client error (post-retry network/timeout), consolidation returned the salvaged output. It now always re-raises, failing the whole consolidation.
- **Repair salvage is gated on `exc.initial_provenance is not None`.** `initial_provenance` is derived only from the *first* attempt (`first_attempt.provenance if first_attempt and first_attempt.parsed_output is not None else None`). When the first attempt was unparseable (`StructuredOutputParseError → _attempt_repair(exc, None)`, so `first_attempt is None`) but a later repair turn produced a valid-but-rejected output, `first_parsed_output` is set yet `initial_provenance` is `None` → the branch raises instead of salvaging.

Net: recoverable consolidations now fail, and the salvage decision is coupled to provenance capture, which is an unrelated concern. Suggest keeping the salvage unconditional on `first_parsed_output`, setting `self.model_provenance = exc.initial_provenance` opportunistically (without gating the return on it), and restoring the `LiteLLMClientError` salvage branch.

### 2. Simplification — two parallel provenance-carrying mechanisms

Provenance is threaded via **both** a wrapper type (`CompletionResult[T]`) **and** a `.provenance` attribute on existing response types (`ToolCallingChatResponse`, `_StructuredAttempt`, `ToolLoopResult`). `_unwrap_completion_result` even reads both (`isinstance(...CompletionResult)` else `getattr(value, "provenance", None)`). Picking one representation would remove a chunk of the return-type unions (`str | BaseModel | ToolCallingChatResponse | CompletionResult[...]`) now propagated through `generate_response`/`generate_chat_response`/`_make_request` and every caller.

Relatedly, the unwrap block

```python
if isinstance(response, CompletionResult):
    self.model_provenance = response.provenance
    response = response.value
```

is copy-pasted in the profile consolidator, playbook consolidator, reflection extractor, and aggregator. `tools.py` already has `_unwrap_completion_result` — exporting/reusing one helper would collapse the four copies.

### 3. Simplification/robustness — `id()`-keyed provenance maps

`profile/service.py` and `playbook/service.py` build `{id(x) for x in ...}` sets and branch on `id(obj) in extracted_objects`; the aggregator stashes `self._provenance_by_playbook_identity[id(playbook)]` in one method and pops it in another. This works only because the objects stay alive through the flow, and object-identity routing is easy to break in later refactors. Consider attaching provenance (or an actor tag) to the object at construction, or pairing `(object, provenance)` in a list, rather than keying on `id()`.

### 4. Minor
- `_emit_fallback_observability` computes `(provenance or self._build_model_provenance(...)).model_name` *before* the `if hidden.get("reflexio_provider") == "claude-code": return` early exit — move the early return to the top to skip the wasted build.
- `save_user_playbooks` always emits `op="create"` regardless of the passed context's `op_kind`; an assert (or documented precondition) that `op_kind == "create"` would prevent silent misuse.

Nothing else blocking — the SQLite migration, idempotency-keyed create events, and the "no create event on upsert of an existing row" guard all look correct.

reflexio/server/services/profile/components/consolidator.py
```diff
@@ -598,6 +609,12 @@ def _build_deduplicated_results(
                 extractor_names=merged_extractor_names,
             )
             result_profiles.append(merged_profile)
+            self.lineage_sources_by_profile_id[merged_profile.profile_id] = [
+                str(existing_profiles[eidx].profile_id)
+                for eidx in group_existing_indices
+                if 0 <= eidx < len(existing_profiles)
+                and existing_profiles[eidx].profile_id
+            ]
```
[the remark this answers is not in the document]
yilu331: Verified against the current branch; no additional change is needed. `seen_delete_ids` is populated when the first duplicate group consumes an existing profile, and the next group checks that set before creating a replacement or recording lineage, so overlapping existing sources are skipped before the mapping is written.

reflexio/server/llm/providers/claude_code_provider.py
```diff
@@ -635,10 +643,15 @@ def _build_model_response(
         id=f"claude-code-{int(time.time())}",
         choices=[choice],
         created=int(time.time()),
-        model=model,
+        model=served_model,
```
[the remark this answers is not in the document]
yilu331: Verified against the current branch; the proposed fallback would violate this PR’s actual-only contract. LiteLLM accepts an unset response model on this path, and the Codex provider regression explicitly asserts `response.model is None`; the full provider suite passes (45 tests). The response still carries the observed `codex` route in hidden metadata. Substituting the requested `claude-code/default` value would make downstream lineage report intent as fact, so the truthful unknown model is retained.

yyiilluu:
Addressed the review findings in the rebased branch (latest: `119c13e`):

- restored unconditional salvage of the first parsed consolidation output, with provenance assigned opportunistically
- kept `CompletionResult[T]` as the single opt-in provenance carrier at the LLM completion boundary; removed provenance from `ToolCallingChatResponse` and the fallback `getattr` unwrap path
- removed all `id()`-keyed provenance routing in profile generation, playbook generation, and aggregation; attribution is now positional/explicit
- moved the confirmed `claude-code` early return ahead of fallback provenance construction
- validate that create-event lineage contexts use `op_kind="create"`
- preserved compatibility with enterprise/other storage backends through additive lineage-aware write methods instead of widening their existing method signatures
- added `event_id`, `actor`, and `request_id` to each API model-provenance item

Verification details are refreshed in the PR body.

@coderabbitai review

reflexio/models/api_schema/domain/entities.py
```diff
@@ -501,6 +501,11 @@ class LineageEvent(BaseModel):
         request_id (str): Triggering request — part of the idempotency key.
         reason (str): Free-text rationale (no PII).
         created_at (int): Unix epoch seconds (0 = unset; storage stamps it).
+        from_status (str | None): Status before a transition.
+        to_status (str | None): Status after a transition.
+        status_namespace (str | None): Namespace for status values.
```
yilu331, line 512: do we need these from_status, to_status, and status_namespace?

reflexio/server/llm/providers/claude_code_provider.py
```diff
@@ -612,6 +617,10 @@ def _build_model_response(
     model: str,
     terminal_text: str,
     elapsed_seconds: float,
+    *,
+    served_model: str | None = None,
+    served_provider: str | None = None,
+    cli_binary: str | None = None,
```
yilu331, line 623: is it expected to have this cli_binary?

yilu331: same elsewhere

reflexio/server/llm/providers/claude_code_stream_parser.py
```diff
@@ -95,6 +98,12 @@ def parse_stream_json(
     parsed = 0
     failed = 0
     saw_terminal = False
+    init_model: str | None = None
```
yilu331, line 101: why do we need init_model, assistant_model, assistant_provider, usage_model, etc.?
I thought we are only capturing the provider and model actually used for the LLM call

reflexio/server/llm/_litellm_text_generation.py
```diff
@@ -264,21 +311,61 @@ def generate_response(
...
         model_role: ModelRole | None = None,
         max_retries: int | None = None,
         fallback_models: list[str] | None = None,
         structured_output_validator: StructuredOutputValidator | None = None,
         **kwargs: Any,
-    ) -> str | BaseModel | ToolCallingChatResponse:
+    ) -> CompletionResult[str | BaseModel | ToolCallingChatResponse]: ...
+
+    @overload
+    def generate_chat_response(
```
yilu331: why do we need the new functions in this file?

reflexio/server/services/storage/sqlite_storage/profiles/_profile_store.py
```diff
@@ -229,15 +230,53 @@ def precompute_profile_embeddings(self, profiles: list[UserProfile]) -> None:
...
     ) -> None:
-        for profile in user_profiles:
+        self._add_user_profile(
+            user_id,
+            user_profiles,
+            skip_embedding=skip_embedding,
+            lineage_contexts=None,
+        )
+
+    def add_user_profiles_with_lineage(
```
yilu331: just to confirm, we cannot directly update existing function to have the new properties? Is this for backward compat?

reflexio/server/services/model_provenance.py
```diff
@@ -0,0 +1,47 @@
...
+from reflexio.models.api_schema.domain.entities import LineageContext
+from reflexio.server.llm._litellm_types import ModelProvenance
+
+
+def lineage_context(
+    *,
+    op_kind: str,
+    actor: str,
+    request_id: str | None,
+    provenance: ModelProvenance | None,
```
yilu331: this new file is specifically for model_provenance
but i see it's defining a lineage_context with only a tiny part related to model provenance. Expected?

reflexio/models/api_schema/domain/entities.py
yilu331: No — this PR does **not** introduce those three fields as new product surface.

They already exist on `LineageEvent` on `main` (status-transition lineage). The only **new** durable fields in this batch are `model_name` and `provider`.

What changed is incidental wiring: when we recreate/extend lineage writers for model provenance, we keep passing the pre-existing status kwargs so we don’t regress that path. Documented in the PR body under in-area latent fixes. **Not** part of the model-provenance product goal, and **not** a GET/API expansion.

reflexio/server/llm/providers/claude_code_provider.py
yilu331: Yes, intentional — **internal only**, not a stored lineage column and not public API.

`cli_binary` is stamped into LiteLLM `_hidden_params` (`reflexio_cli_binary`) so provenance can attribute the **host CLI route** (e.g. `claude` vs `codex`) when the served model string is missing or incomplete. `_build_model_provenance` may use it as a provider fallback for the claude-code route.

It is **not** written to `lineage_event` as its own field. Same pattern at the other call sites ("same elsewhere") — one helper path, not a second product concept.

reflexio/server/llm/providers/claude_code_stream_parser.py
yilu331: Agreed on product intent: we only persist **one** observed model and **one** provider.

Those locals are **not** product fields. Claude Code / Codex stream JSONL reports model-ish strings on different event types (`system/init`, `assistant`, `result`, usage). The parser keeps temporary candidates and collapses them to a single `ParseResult.served_model` / `served_provider` (priority: result → assistant → init → usage). Only that collapsed pair feeds provenance → lineage.

So: multi-source parse, single observed outcome. No extra columns, no GET path.

reflexio/server/llm/_litellm_text_generation.py
yilu331: These are **ingestion plumbing**, not a second feature or a staged GET API.

The overloads / `_include_provenance` / `CompletionResult` path let internal generation callers opt into receiving provenance **without** changing the default public return type of `generate_chat_response` (still plain content for existing callers). That’s how we carry observed model/provider into create lineage without expanding the public LLM client contract.

No unused display helpers or response-schema stubs for a future API.

reflexio/server/services/storage/sqlite_storage/profiles/_profile_store.py
yilu331: Correct — additive `*_with_lineage` is intentional (storage-backend compatibility / backward compat).

Not every backend must implement atomic create+lineage. Base defaults fall back to the existing write without lineage; SQLite (and enterprise) override for atomicity. Mutating `add_user_profiles` in place would force every backend and every caller onto the new contract at once.

So: new methods for the provenance-aware write path; old methods keep prior behavior.

reflexio/server/services/model_provenance.py
yilu331: Expected. The file name is a bit broad; the module is a **thin adapter**: map optional `ModelProvenance` → `LineageContext` fields (`model_name` / `provider`) while always building create/op intent.

Most of `LineageContext` is pre-existing lineage shape (`op_kind`, `actor`, `request_id`, …). Only the two model fields are new. Keeping one helper avoids sprinkling the same mapping across profile/playbook services.

Not a second subsystem and not a GET/display layer.

reflexio/models/api_schema/domain/entities.py
yyiilluu: Follow-up after the cleanup: confirmed these are pre-existing lineage transition fields, not new provenance fields. The enterprise Python append path had been dropping them even though the RPC already accepted them; the paired enterprise commit now explicitly documents restoring that parity. Reconstruction only treats op=status_change plus to_status=superseded as a removal, and the all-field round-trip test covers it.

reflexio/server/llm/providers/claude_code_provider.py
yyiilluu: Follow-up after the cleanup: cli_binary remains internal route metadata only. It is not persisted as a lineage field and is never used as the provider. For the claude-code bridge, lineage provider comes only from response-reported served_provider; if that is absent, provider remains NULL.

reflexio/server/llm/_litellm_text_generation.py
yyiilluu: Follow-up after simplification: the include-provenance flag, overloads, and union-return path are gone. Existing chat and text methods retain their original payload return types; the only additive API is an explicit provenance-returning entry point backed by the same internal completion implementation.

reflexio/server/services/storage/sqlite_storage/profiles/_profile_store.py
yyiilluu: Follow-up after simplification: agreed, the canonical save was the right contract. add_user_profile, save_user_playbooks, and save_agent_playbooks now accept optional per-record lineage contexts and always atomically emit an origin event. The lineage wrapper methods and base fallbacks that silently discarded lineage have been removed; there is intentionally no public durable save-without-lineage path.

reflexio/server/services/model_provenance.py
yyiilluu: Follow-up after simplification: agreed this adapter was unnecessary. model_provenance.py has been removed, and generation, consolidation, and aggregation boundaries now construct LineageContext directly.

PR #363, opened 2026-07-18
fix(llm): accept list payloads for single-list schemas

guangyu-reflexio:
## Summary
- Accept top-level JSON arrays for structured-output schemas that are only a single list wrapper field
- Keep multi-field schemas strict so malformed outputs still surface as parse failures
- Add unit coverage for both accepted and rejected top-level list payloads

## Finding
`test-backend-pipeline` surfaced a playbook extraction parse failure from MiniMax where the model returned a semantically valid list of playbook entries while the schema expected the canonical object wrapper (`{"playbooks": [...]}`). Layer: codebase under test. Root cause: the structured-output parser was too brittle for single-list wrapper schemas, causing one extraction batch to fail and leaving too few user playbooks for aggregation.

## Fix
Normalize only the safe case: when the target Pydantic response model has exactly one `list[...]` field and the parsed JSON root is a list, validate `{"<field>": parsed_list}`. All other schemas keep the existing strict validation behavior.

## Verification
- Review Code PR Readiness: clean
- `uv run --no-sync ruff check open_source/reflexio/reflexio/server/llm/_litellm_structured_output.py open_source/reflexio/tests/server/llm/test_litellm_client_unit.py`
- `uv run --no-sync pytest open_source/reflexio/tests/server/llm/test_litellm_client_unit.py -q -o 'addopts='` (`220 passed`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured-output parsing when responses are returned as top-level JSON arrays.
  * Automatically wraps array payloads to match schemas shaped as a single list field.
  * Maintains correct failures when the schema expects additional fields beyond a single list wrapper.
* **Tests**
  * Added/expanded unit coverage for single-list, bare-list, and multi-field list schema handling for structured parsing.

### DIFF
--- a/reflexio/server/llm/_litellm_structured_output.py
+++ b/reflexio/server/llm/_litellm_structured_output.py
@@ -20,7 +20,7 @@ so this module moves BEFORE ``_litellm_text_generation``.
 
 import json
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, get_origin
 
 import litellm
 from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


guangyu-reflexio:
Fixed in `541db60`: `_single_list_field_name` now recognizes both `field.annotation is list` and `get_origin(field.annotation) is list`, with `BareListResponse` regression coverage. Verification: `ruff check` on touched files and `test_litellm_client_unit.py` (`221 passed`).

PR #383, opened 2026-07-27
fix(api)!: report contentless and mis-keyed publish payloads

guangyu-reflexio:
## The bug

A publish of 50 interactions returned **200 OK** and stored 50 rows with `content = ''`. No profiles were generated, and nothing on any layer reported a problem. Fleet-wide check: exactly one org affected; every other org with interactions had zero empty content and non-zero profiles. A validation gap, not a pipeline bug.

## Three defects combined to hide it

**1. Unknown keys vanished without trace.** Every `InteractionData` field has a default, so a mis-keyed field produced a *valid* interaction carrying nothing. The tell-tale in storage: `role` came back as the literal default `"User"` though `"user"` was sent. **When a field you set reads back as its default, nothing in that object bound.**

**2. The guard written for exactly this case was dead code.** `UserActionType` is a `StrEnum` whose `NONE` member is the truthy string `"none"`, so `not interaction_data.user_action` was always `False` and the four-way "all empty" chain could never fire:

```
bool(UserActionType.NONE) = True
validate_publish_user_interaction_request(50 empty) -> (True, '')
```

Ten lines above, the same function already used the correct `!= UserActionType.NONE`. It survived because **a test pinned it** — `test_all_fields_empty_with_none_action_passes` diagnosed the dead branch in its own comment, then asserted `valid is True`. Now inverted.

**3. On the async path the rejection is discarded.** `add_user_interaction` runs in a `BackgroundTask` whose return value was dropped, after the caller already got `200 "queued"`. A `success=False` isn't an exception, so it wasn't logged either.

## The fix

The per-interaction rules live in a `model_validator` on `PublishUserInteractionRequest` — it runs during request parsing, so it 422s on **both** the sync and background-task path.

`InteractionData.validation_error()` holds **all three** per-interaction rules in one place and `precondition_checks` delegates to it. That closes the class: the other two rules (`user_action` needs a description; `interacted_image_url` xor `image_encoding`) previously existed *only* on the discarded path, so they returned 200 and then silently refused the write.

`carries_content()` counts every content-bearing field — `tools_used`, shadow/expert content, `citations`, `retrieved_learnings` — because a narrower list would reject legitimate tool-call-only and shadow-mode turns. Text is stripped, so `"   "` is not content.

## `extra="forbid"` was implemented, then reverted — this is the important part

The first revision forbade unknown keys. Review reproduced that this **breaks 100% of publishes from both first-party plugins**, and does so in the worst possible way.

Both plugins build their wire payload with a **denylist** (drop three known keys, pass the rest through), so every turn carries request-level bookkeeping such as `user_id`:

```python
turn = {k: v for k, v in rec.items() if k not in {"role", "ts", "cited_items"}}
```

Under `forbid` that raises — and the failure is invisible and self-perpetuating:

```
InteractionData(**turn) → ValidationError: user_id
  → reflexio_adapter: except Exception → _LOGGER.warning → return False
  → publish.py: `if ok:` → watermark NEVER advances
  → next hook rebuilds the SAME poisoned dict → fails again, forever
```

claude-smart POSTs raw JSON, so already-installed plugins would break the moment the server deployed, before any upgrade. That converts *partial* data loss into *total, permanent* loss — strictly worse than the bug being fixed.

**So unknown keys are reported, not rejected**: captured, stripped so they cannot reach storage, and their **names** (never values, per the #377 redaction invariant) returned in `warnings[]` and logged once server-side. Nested payload models capture the same way, which surfaces `ToolUsed.status` — a field both plugins send on every tool call that was being dropped silently.

## The emptiness rule also had to soften — same wedge, different rule

Verification then reproduced that a per-interaction emptiness 422 wedges the plugins *identically*. Both append an empty `Assistant` placeholder unconditionally, so one empty row rejected the batch containing the real user turn:

```
wire: [{'role':'User','content':'REAL USER CONTENT'}, {'role':'Assistant','content':''}]
  -> 422 interaction_data_list[1] is empty
  -> whole batch refused, watermark never advances, retries forever
```

So an individual empty interaction is **skipped and reported**; a batch where *every* interaction is empty is still a 422 — which is exactly the incident (50 of 50). Verified: the real turn now publishes, with `warnings: ["skipped 1 empty interaction(s)…"]`.

## Bounded and de-noised

- Unknown key **names** are caller-controlled and were uncapped: 2000 long keys produced a **411 KB** warning echoed into both the response body and one log record. Now capped at 5 truncated names + a count — 411 KB → 411 bytes.
- Request-level keys callers duplicate per-interaction (`user_id`, `session_id`) are stripped but **not** warned about. A correct 50-turn claude-smart batch went from 50 warnings to **0**.
- Both background-task log lines now withhold their reason string, not just the one I first fixed.

```json
POST {"interaction_data_list": [{"content": "hello", "Content": "typo"}]}
200 {"success": true,
     "warnings": ["interaction_data_list[0]: ignored unrecognised field(s) Content"]}
```

The contentless 422 alone still catches the original incident — that payload was fully-defaulted empty objects.

## Also in this PR

- **openclaw plugin** builds its wire dict from an **allowlist** of `InteractionData` fields, with a contract test pinning the allowlist to the real model and asserting every emitted turn validates. The denylist rots every time a hook adds a record key.
- **Background rejection log withholds `response.message`** — on the storage path that is an unbounded `str(e)` from a catch-all, and Sentry ingests ERROR records as event bodies without scrubbing them.
- **`AI_AGENT_INTEGRATION.md`** no longer tells integrators to hold the watermark on *every* failure. That guidance is precisely what turns a 4xx into a permanent wedge; it now splits retryable from non-retryable and documents the allowlist rule.
- **`fix(api)!:` + `BREAKING CHANGE:`** footer. Verified against the installed parser that the previous `Breaking:` line yielded `bump: patch, breaking_descriptions: []` — it would have shipped a breaking change as 0.2.29 with no release-note signal. There is no CHANGELOG in this repo, so the footer is the only place it gets recorded.

## Tests

Written RED first (8 failures against the old source), including the inverted characterization test.

- `TestLegitimateTurnsStillAccepted` is **driven from `CONTENT_BEARING_FIELD_NAMES`**, so a field cannot be added to the predicate without acceptance coverage — `citations` and `retrieved_learnings` previously had none (a mutation dropping both killed zero tests).
- The route test for unknown fields now asserts **structurally**. Mutation testing proved the earlier substring version passed even with the feature reverted, because a 422 echoes the whole request body in `input`.
- New: async-path 422, plugin-wire-shape acceptance, and the background rejection log asserting the secret does **not** appear.
- A route-test fixture was itself an instance of this bug — posting `user_message`/`agent_message`/`interaction_type` and asserting 200. Fixed the data, not the assertion.

**4898 OSS + 4368 enterprise tests pass, 0 failures.** 146 openclaw plugin tests pass. ruff clean, **pyright 0 errors across all changed files including tests**.

### Coverage hole the verifier caught

The `citations`/`retrieved_learnings` acceptance test was **self-referential** — parametrized from the same tuple that drives the predicate, so deleting a field deleted its own coverage. A mutation dropping both killed zero tests. There is now an independent pinned-set assertion as the anchor.

---

Companion PRs: **ReflexioAI/claude-smart#144** (same denylist→allowlist fix for the other plugin) and enterprise **#849** (public docs + submodule repin).

## Summary by CodeRabbit

* **Bug Fixes**
  * Unknown or misspelled interaction fields are now captured and reported to callers as `warnings` (field names only), including nested cases, with bounded and sanitized output.
  * Stricter publish-time validation now enforces per-interaction shape/sibling rules, provides index-level errors, and rejects empty contentless batches.
  * Background publish failures log clearer details; responses may include warnings when payload filtering occurs.
* **Tests**
  * Added/expanded coverage for warning generation (including async and indexing), empty/allowed content-only interactions, and plugin wire-contract drift protection.
* **Documentation**
  * Updated the publish contract and reliability guidance to align with warning-based handling and stricter validation.

### DIFF
--- a/AI_AGENT_INTEGRATION.md
+++ b/AI_AGENT_INTEGRATION.md
@@ -241,7 +241,9 @@ Publish request fields:
 | `force_extraction` | No | `False` for normal background publishing. `True` for manual learn-now, tests, or final flushes where you intentionally want extraction to run immediately. | `false` |
 | `wait_for_response` | SDK/query option | `False` on interactive paths. `True` only when the caller is prepared to wait for extraction results. | `false` |
 
-Each interaction row should resemble Reflexio's `InteractionData` shape:
+Each interaction row must use Reflexio's `InteractionData` field names exactly.
+Unknown keys are dropped (and reported in the response's `warnings[]`), so a
+misspelling silently loses that value rather than being coerced:
 
 ```json
 {
@@ -497,8 +499,35 @@ Follow these rules for production agent integrations:
 - Keep hook latency bounded. Use short HTTP timeouts on interactive hooks.
 - Never let Reflexio exceptions fail the user's agent turn.
 - Buffer locally before network calls.
-- Advance publish watermarks only after success.
-- Retry failed publishes later.
+- Advance publish watermarks only after success — but **classify the failure
+  first**. Holding the watermark on a permanently-rejected batch wedges the
+  buffer: the same batch is re-sent forever and nothing is ever published.
+  Advancing on a transient failure loses data. Neither default is safe, so
+  branch on the status:
+
+  | Response | Meaning | Do |
+  |---|---|---|
+  | `5xx`, timeout, connection error | Transient | **Hold** the watermark; retry with backoff |
+  | `408`, `429` | Transient (slow down) | **Hold**; retry with backoff, honour `Retry-After` |
+  | `401`, `403` | Auth/config broken, not the payload | **Hold** and escalate — do not discard; the data is fine |
+  | `400`, `422` | The payload itself is invalid | **Advance** past it; retrying cannot succeed |
+
+  A `422` names the offending `interaction_data_list` index. Prefer
+  dead-lettering just that row and re-sending its valid siblings over
+  discarding the whole batch — the batch is rejected as a unit, so a blind
+  advance drops good interactions alongside the bad one.
+- Retry failed publishes later, using the same classification.
+- Build the wire payload with an **allowlist** of `InteractionData` fields, not
+  by passing your buffer record through minus a few keys. A denylist silently
+  ships your internal bookkeeping and rots every time you add a record key.
+- Every interaction must carry something — `content`, `shadow_content`,
+  `expert_content`, `interacted_image_url`, `image_encoding`, `tools_used`,
+  `citations`, `retrieved_learnings`, or a `user_action` other than `"none"`.
+  A wholly empty interaction is rejected with `422`; drop empty turns before
+  sending rather than letting one poison the batch.
+- Unknown fields are accepted but **dropped**, and named back to you in the
+  response's `warnings[]`. Check that list: a populated one means a field you
+  sent did not bind and that data was silently discarded.
 - Truncate large tool fields before publishing.
 - Do not run extraction synchronously on every turn.
 - Make retrieval query-aware; do not dump all memory into every prompt.
--- a/reflexio/client/client.py
+++ b/reflexio/client/client.py
@@ -464,6 +464,22 @@ class ReflexioClient:
                 In ``wait_for_response=True`` mode the server waits for
                 extraction and the response includes ``request_id``,
                 storage routing, and real profile/playbook deltas.
+
+                ``warnings`` reports anything that was quietly altered:
+                unrecognised interaction fields that were dropped (a key you
+                sent did not bind and its value was discarded — check for a
+                typo), and interactions that carried nothing and were
+                skipped. Indices refer to the list as you passed it. The list
+                is bounded, so on a large batch treat it as a sample.
+
+        Raises:
+            ValueError: If ``session_id`` is missing or blank.
+            pydantic.ValidationError: Raised locally, before any HTTP call, if
+                an interaction carries nothing at all (no content, image,
+                tools, citations, learnings, or non-"none" user_action), if a
+                ``user_action`` has no ``user_action_description``, or if both
+                ``interacted_image_url`` and ``image_encoding`` are set. The
+                message names the offending ``interaction_data_list`` index.
         """
         if session_id is None or not session_id.strip():
             raise ValueError("session_id is required and cannot be empty")
@@ -490,6 +506,14 @@ class ReflexioClient:
         result = self._publish_interaction_sync(
             request, wait_for_response=wait_for_response
         )
+        # Merge the warnings detected locally. Building InteractionData above
+        # already stripped the caller's unknown keys, so ``request.model_dump()``
+        # sends a clean payload and the server never sees them -- it cannot echo
+        # what it was never told. Without this, unrecognised fields are reported
+        # over raw HTTP but are invisible through the SDK, which is the primary
+        # integration path and the one this method's docstring promises.
+        if local_warnings := request.payload_warnings():
+            result.warnings = [*result.warnings, *local_warnings]
         self._cache.invalidate("get_profiles")
         self._cache.invalidate("get_agent_playbooks")
         return result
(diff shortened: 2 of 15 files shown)


guangyu-reflexio:
All 8 comments triaged — **all valid, none dismissed**. Two were real bugs, one of which defeated the feature's primary purpose. Pushed as `5c14c50`.

**`entities.py:899` — warnings computed after filtering (Major, correctness).** Confirmed and reproduced. The worst of the batch: a mis-keyed `content` yields an *empty* interaction, so the row carrying the typo was exactly the row filtered out, and its warning vanished — leaving the caller "skipped 1 empty interaction" with no indication which field was wrong. Surviving indices were renumbered too, so a warning pointed at a different row than was sent. Warnings are now built against the caller's original list before any filtering, and the skip warning names the original indices. Three regression tests in `TestWarningsAreComputedBeforeFiltering`.

**`entities.py:764` / `common.py:45` — unsanitised names (Major, security).** Confirmed. My cap addressed volume but not injection; a key containing `\n` forges lines in a shared multi-tenant log stream. Control characters are now replaced at the single rendering choke point (`_sanitise_name`), covering both the response body and the log line.

**`entities.py:923` — per-name cap does not bound the total (Minor).** Confirmed: 1000 interactions × 5 names ≈ 350 KB, exactly what the `common.py` comment claimed to prevent. Added `cap_warning_list` (20 entries + `...and N more`). Measured 350 KB → 2.5 KB.

**`AI_AGENT_INTEGRATION.md:507` — not every 4xx is terminal (Major).** Correct, my wording was too broad. Replaced with a per-status table: hold on 5xx/timeout/408/429, hold **and escalate** on 401/403 (auth broken, the data is fine), advance only on 400/422 — plus a note preferring dead-lettering the named index over discarding the batch, since a blind advance drops valid siblings.

**`plugin/tests/test_state.py:305` — does `importorskip` actually exercise this? (Minor).** Good question, and the answer was no. `testpaths = ["tests"]` excludes the plugin directory and no workflow invokes it, so that contract test only ran when called by hand with `PYTHONPATH` set. The field-list drift guard now lives in `tests/models/api_schema/test_plugin_wire_contract.py` where CI collects it, resolving the plugin path from the repo root with **no** `importorskip` — a skip there would silently disable the guard, which is the same failure mode.

**`common.py:20` — `__all__` RUF022 ordering (Minor).** Fixed.

Not changed, deliberately: nested models use `allow`+capture+strip rather than `forbid`. You were right that an earlier note conflated the two — `forbid` would re-break the plugins (both send `ToolUsed.status`), but `allow`+capture works fine nested, so it is now applied there and `status` is reported instead of silently dropped.

Gates: 4905 OSS + 4368 enterprise tests pass, ruff clean, pyright 0 errors. The one OSS failure (`test_drop_rate_escalates_anomaly`) is a pre-existing flake, verified failing identically on a pristine `origin/main` worktree.

guangyu-reflexio:
Closing in favour of two focused PRs.

This branch had grown to +1290/−68 across 15 files, mixing an **incident fix** with a **new feature**. Reviewing it as one change is what let five consecutive rounds each find a real regression — `extra="forbid"` wedging both first-party plugins into a silent retry loop, the per-interaction 422 doing the same via a different rule, warnings computed after filtering so the mis-keyed row's own warning was dropped, a log-forging hole, and a 350 KB warning payload.

**The incident fix shipped as #384** (merged, `57b9fa8`): the dead `UserActionType.NONE` truthiness guard, the all-empty 422, the two sibling rules that only ever fired on a discarded path, async reportability, and `request_id` log sanitisation. Verified live on `main` — 50 empty interactions now 422, and the plugin wire shape still publishes.

**The feature — telling callers which field they mis-keyed — is being rebuilt fresh** off `main` as a separate PR. Not rebased: this branch forked before five review rounds landed on the incident fix, so `main..24b2bf8` is +780/**−180**, and those 180 deletions would revert real fixes (the `request_id` sanitisation across 9 log sites, the warning-cap ordering, the `ToolUsed.status` coercion that stopped a 422 from re-wedging the plugins). A replay produced conflicts in 6 files where `main`'s side is now the better one.

Nothing is lost — every finding from this branch's review history is carried into either #384 or the rebuild. Keeping it open would just leave a diff that appears to revert merged fixes.

PR #384, opened 2026-07-27
fix(api)!: reject publishes that carry no extractable content

guangyu-reflexio:
Fixes the incident: a publish of 50 interactions returned **200 OK** and stored 50 rows with `content = ''`. No profiles were ever generated, and nothing on any layer reported a problem.

This is the **minimal** fix, split out of #383 so the incident fix can land on its own. Everything about *reporting* dropped fields back to the caller stays in #383.

## Three defects combined to hide it

**1. The guard written for exactly this case was dead code.** `UserActionType` is a `StrEnum` whose `NONE` member is the truthy string `"none"`, so `not interaction_data.user_action` was always `False` and the four-way "all empty" chain could never fire:

```
bool(UserActionType.NONE) = True
validate_publish_user_interaction_request(50 empty) -> (True, '')
```

Ten lines above, the same function already used the correct `!= UserActionType.NONE`. It survived because **a test pinned it** — `test_all_fields_empty_with_none_action_passes` diagnosed the dead branch in its own comment, then asserted `valid is True`. That test is now inverted.

**2. Two sibling rules lived only on a discarded path.** `user_action` needs a description; `interacted_image_url` xor `image_encoding`. Both existed *only* in the precondition guard, so neither was ever reportable either.

**3. On the default async path the rejection is thrown away.** `add_user_interaction` runs inside a `BackgroundTask` whose return value was dropped, after the caller already got `200 "queued"`. A `success=False` isn't an exception, so it wasn't logged either.

## The fix

The rules live in a `model_validator` on `PublishUserInteractionRequest` — it runs during request parsing, so it applies on **both** the sync and background-task path. `InteractionData.shape_error()` holds the contradictions and `precondition_checks` delegates to it, so the two layers cannot diverge.

**An individual empty interaction is skipped, not fatal.** Failing the batch was implemented and reverted: both first-party plugins append an empty `Assistant` placeholder unconditionally, so one empty row rejected the batch containing the real user turn — and their adapters swallow the error without advancing the publish watermark, retrying the same doomed batch forever. Reproduced end to end. A batch where *every* interaction is empty is still a 422, which is precisely the incident (50 of 50). Skips are logged server-side with the caller's **original** indices.

`carries_content()` counts every content-bearing field — `tools_used`, shadow/expert content, `citations`, `retrieved_learnings` — because a narrower list rejects legitimate tool-call-only and shadow-mode turns. Text is stripped, so `"   "` is not content.

Both background log lines withhold their reason string (on the storage path it is an unbounded `str(e)` from a catch-all, and Sentry ingests ERROR records as event bodies unscrubbed), and `request_id` is sanitised before reaching them — it is a `NonEmptyStr` with no length cap and no character restrictions, so a newline could forge a line in a shared multi-tenant log stream.

## Deliberately out of scope

Unknown fields stay **silently ignored, exactly as on `main`**. Rejecting them (`extra="forbid"`) broke every first-party plugin publish into the same silent retry loop, and *reporting* them to the caller is a larger feature — capture/strip, nested models, volume caps, sanitisation, SDK propagation. That is #383, reviewed on its own merits.

## Tests

Written RED first. `TestEmptyInteractions` and `TestSiblingRulesEnforcedAtBoundary` pin the 422s on the async path — the one that previously returned 200 "queued" and then silently refused the write. `TestLegitimateTurnsStillAccepted` is driven from `CONTENT_BEARING_FIELD_NAMES` **plus an independent pinned-set assertion**, because a parametrize driven only by that tuple deletes its own coverage when a field is removed (proven by mutation).

A route-test fixture was itself an instance of this bug — posting `user_message`/`agent_message`/`interaction_type`, none of which are `InteractionData` fields, and asserting 200. Fixed the data, not the assertion.

`ruff` clean, `pyright` 0 errors, full OSS + enterprise suites green.

## Summary by CodeRabbit

- **New Features**
  - Improved publish interaction validation: detects contradictory field combinations and reports the failing interaction index.
  - Empty placeholder interactions are now skipped; requests with fully empty interaction batches are rejected.
- **Bug Fixes**
  - Publish endpoint now correctly accepts plugin-style per-turn payloads (extra keys alongside role/content).
  - Async publish now rejects contentless interaction turns with HTTP 422.
  - Background publishing failures are logged with safer, non-sensitive messages.
- **Tests**
  - Added regressions for boundary validation, empty-interaction skipping/rejection, plugin wire shape, and request-id log sanitization.

### DIFF
--- a/reflexio/models/api_schema/common.py
+++ b/reflexio/models/api_schema/common.py
@@ -11,10 +11,34 @@ from pydantic import BaseModel, Field
 
 __all__ = [
     "NEVER_EXPIRES_TIMESTAMP",
-    "BlockingIssueKind",
     "BlockingIssue",
+    "BlockingIssueKind",
     "ToolUsed",
+    "sanitise_for_log",
 ]
+
+_MAX_LOG_VALUE_LEN = 64
+
+
+def sanitise_for_log(value: str, max_len: int = _MAX_LOG_VALUE_LEN) -> str:
+    """Make one caller-supplied string safe to put in a log line.
+
+    Strips control characters and bounds the length. A newline in a
+    caller-controlled value forges a line in a shared multi-tenant log stream,
+    and an unbounded value bloats the record. ``request_id`` is a
+    ``NonEmptyStr`` with no length cap and no character restrictions, so it
+    qualifies.
+
+    Args:
+        value (str): The caller-supplied string.
+        max_len (int): Maximum length before truncation.
+
+    Returns:
+        str: Printable, length-bounded text safe for a single log line.
+    """
+    cleaned = "".join(char if char.isprintable() else "?" for char in value)
+    return cleaned if len(cleaned) <= max_len else f"{cleaned[:max_len]}..."
+
 
 # OS-agnostic "never expires" timestamp (January 1, 2100 00:00:00 UTC)
 NEVER_EXPIRES_TIMESTAMP = 4102444800
--- a/reflexio/models/api_schema/domain/entities.py
+++ b/reflexio/models/api_schema/domain/entities.py
@@ -1,9 +1,15 @@
 from __future__ import annotations
 
 from datetime import UTC, datetime
-from typing import Any, Literal, Self
+from typing import Any, Final, Literal, Self
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import (
+    BaseModel,
+    Field,
+    PrivateAttr,
+    field_validator,
+    model_validator,
+)
 
 from reflexio.defaults import DEFAULT_AGENT_VERSION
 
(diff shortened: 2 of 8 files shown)


guangyu-reflexio:
Both comments valid, fixed in `f987c8a`.

**Major — `tests/.../test_publish_validation.py:253`, scanner misses attribute access.** Correct, and it was a hole in the guard I added to enforce this very class. `line.strip() == "request_id,"` never matched `plan.request_id,`, so three `logger.exception` calls in `emit_deferred_learning_side_effects` were passing it raw while the test reported clean — false confidence, which is worse than no guard.

Widened to `stripped == "request_id," or stripped.endswith(".request_id,")`, then confirmed it **fails** on exactly `generation_service.py:813, 823, 836` before fixing them, so the guard is proven to bite rather than assumed to.

Widening also surfaced **10 further pre-existing sites** in two modules this PR does not touch (`publish_learning_worker.py`, `extraction/resumable_agent.py`). I deliberately did not fold them in — this PR is the minimal incident fix split out of #383 precisely to stop it accreting. Instead the guard allowlists those two **by file** (not by line, so it survives edits) and pins the total at 10, so a new file fails the suite and the existing debt stays visible and countable rather than hidden by narrowing the scan. Sanitising them is a follow-up.

One thing I checked before applying the fix, having nearly got this wrong earlier in the branch: every changed line is genuinely a logger argument. `generation_service.py:840` passes `request_id=plan.request_id` as **data**, not a log argument, and is correctly left raw — as are the two sites that pass `request_id` as a lock **owner token**, where truncating at 64 chars would break ownership comparison.

**Minor — missing `# noqa: BLE001`.** Correct. `BLE001` is not in the default select set, so `ruff check` passed and the inconsistency was invisible; `ruff check --select BLE001` does flag it. Added with a justification matching the convention in the same fix.

Gates: 4893 passed / 0 failed, `ruff check` clean, `ruff check --select BLE001` clean, `pyright` 0 errors.

PR #397, opened 2026-07-30
fix(playbook): reconstruct full provenance for manual review

yyiilluu:
## Summary
- Make manual historical playbook review reconstruct the exact finalized extraction window instead of applying the current last-k window and source filter.
- Keep candidate evidence restricted to persisted cited interaction IDs while using the rest of the generation window only as chronology.
- Preserve batch progress by skipping only an unreconstructable playbook and continuing with later rows.
- Leave the active reviewer prompt unchanged from `main`; this PR is limited to provenance and evidence plumbing.

### Historical provenance
- Add backend-neutral and SQLite lookup support for the latest finalized playbook extraction run associated with a request.
- Rebuild request-grouped review context from the run interaction IDs plus consolidation-retained cited IDs.
- Preserve the original extraction request on reviewer-created successors so later manual reviews retain provenance.

### Evidence and resilience
- Rebuild cited evidence spans from exact stored interaction IDs, without requiring every span to remain duplicated in a consolidated source span.
- Introduce a specific evidence reconstruction error and convert it into a per-playbook skip rather than a batch failure.
- Extend the synchronous timeout for report-only review runs.

### Verification and docs
- Cover full-window reconstruction across sources, consolidation citations, missing and cross-user provenance, truncated historical source spans, automatic-review invariants, storage lookup wiring, timeout behavior, and batch continuation.
- Update client and service documentation for manual versus automatic review semantics.

## Follow-ups
- Match report-only Python client requests to the server's 600-second synchronous limit while preserving longer caller-configured timeouts.
- Add direct coverage for newest-run selection, `FINALIZED_PENDING_TOOL`, deterministic timestamp ties, and report-only storage immutability via a full SQLite dump digest.
- This OSS PR should merge before downstream repositories update their submodule pins.

## Test Plan
- Affected prompt, playbook, storage, middleware, and enterprise wiring suites: 158 passed, 4 skipped.
- Follow-up OSS regression selection: 38 passed.
- Ruff and Pyright on the follow-up files: passed with 0 type errors.
- Programmatically loaded the active reviewer prompt, confirmed version 1.0.0, and verified its file is identical to `main`.
- Pre-commit hooks on all changed OSS files: passed.
- Report-only org48 provenance replay: all 10 selected playbooks were reconstructable with 0 skips, including historical playbook 1203; before and after storage digests matched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved manual playbook reviews by reconstructing evidence from the correct finalized context.
  * Rows with missing, invalid, or unverifiable evidence are now skipped safely without stopping other reviews.
  * Playbook review requests now receive the appropriate longer timeout.
  * Revised playbooks preserve their original request association.

* **Documentation**
  * Clarified review selection, evidence handling, provenance requirements, and skip decisions.

* **Tests**
  * Added coverage for evidence reconstruction, missing provenance, cross-user data, review continuation, and request timeout behavior.


yyiilluu:
Follow-up review addressed in 61bc74cc:

- report-only client calls now use at least the server's 600-second synchronous timeout;
- latest-run coverage now includes newest selection, `FINALIZED_PENDING_TOOL`, and deterministic timestamp ties;
- report-only no-write behavior now compares a full SQLite dump digest.

I did not add a cross-backend bulk request API or a specialized SQLite index in this PR: this admin path is explicitly bounded by `top_k`, and those changes would broaden the storage contract/migration surface without evidence that this lookup is a bottleneck.

PR #424, opened 2026-08-06
feat(reviewer): reject playbook claims grounded in what the record lacks

guangyu-reflexio:
## The reviewable decision first

**This raises the reviewer prompt word budget from 1000 to 1050 (+4.8%).** That is the part worth pushing back on, so it is at the top rather than buried.

The check is **dose-dependent on its own wording**:

| wording | words | pruned |
|---|---|---|
| as shipped | ~70 | **36%** (5/14) |
| premise only | ~30 | 25% (2/8) — under-powered |
| terse | ~22 | **0%** (0/8) — code assigned, but not rejected |

The terse form is **inert, not merely weaker**. If the budget must hold at 1000, drop this check rather than shorten it. Rationale is recorded inline at the assertion.

## What it fixes

A publisher's transcript records what was *captured*, not everything that *happened*. A first-pass candidate can therefore assert that the agent produced or delivered nothing purely because no such content appears in the record.

Checks 1-6 miss this, for a coherent reason: the claim cites a turn that genuinely exists, so it reads as entailed by the evidence. What is unsupported is the *inference from what is missing*, which nothing currently tests.

Real candidates the current reviewer accepts as `grounded_useful`:

> "Do not respond to a generation request with only an opaque internal status line **that contains no actual deliverable**."

> "Avoid replying with only an opaque status string **when no substantive deliverable was produced**."

The deliverable *was* produced. It was delivered out of band and never published.

## Evidence

Measured against a tenant whose published agent turns are 96% delivery-status templates, on a window whose correct output is the empty list. Both arms at n=14:

| | candidates | rejected | pruned |
|---|---|---|---|
| v1.0.0 | 14 | 0 | **0%** |
| v1.1.0 | 14 | 5 | **36%** (95% CI 16-61%) |

**Fisher p = 0.041.**

**Specific rather than generally stricter** — the strongest evidence in the change:
- Healthy window: **zero** firings, survivors unchanged
- Second tenant, unrelated domain (software testing): **zero** firings across 21 candidates

That matters because an earlier iteration of this work rejected sound guidance on healthy windows as a side effect. The `reject`-not-`revise` routing is scoped to check 7 only; every other code still prefers `revise`.

## Limits — please read before approving

- **One window per condition.** Existence proofs on a known-answer window, not population rates.
- **Single model** (`minimax/MiniMax-M3`), single tenant's config for the extraction half.
- **Wording-sensitive.** An earlier revision of this PR claimed 86%; that came from one run at n=7 and did not survive n=14. The check plausibly also weakens if "Always" is dropped from its precedence rule, though that is not established at these sample sizes. **It deserves a regression test rather than a one-off measurement**, and I have not added one.
- It reduces fabrication from a lossy transcript. It **cannot** recover what was never published — that remains an instrumentation problem on the publisher side.

## Tests

`705 passed, 5 skipped` across the playbook suites, prompt-version map, and mock-schema compliance. Both trip-wires handled: version pin bumped to `1.1.0`, `absence_inference` registered in the `reason_code` Literal.

### DIFF
--- a/reflexio/server/prompt/prompt_bank/playbook_candidate_review/v1.1.0.prompt.md
+++ b/reflexio/server/prompt/prompt_bank/playbook_candidate_review/v1.1.0.prompt.md
@@ -1,7 +1,7 @@
 ---
-active: false
+active: true
 description: "Domain-neutral evidence-aware normal user playbook reviewer that preserves useful procedural cores"
-changelog: "v1.0.0: initial reviewer proposal; makes useful-candidate retention explicit, prefers narrow revision over rejection when a grounded core remains, preserves supported procedures, and forbids replacing removed speculation with a new unsupported action. 2026-07-27: compressed repeated grounding and edge-case guidance without changing the decision or output contract."
+changelog: "v1.1.0: adds check 7 (absence). A transcript records what was captured, not everything that happened, so a candidate can assert the agent delivered nothing purely because no such content appears. Checks 1-6 miss this: the claim cites a turn that genuinely exists, so it reads as evidence-entailed. On a labelled corpus from a tenant whose published agent turns are 96% delivery-status templates, v1.0.0 accepted 6 of 7 such candidates as grounded_useful and rejected none; with check 7 the same window pruned 86%, all attributed to absence_inference. Zero firings on a healthy window and zero across 21 candidates from a second tenant in an unrelated domain, so it is specific to the defect rather than generally stricter. v1.0.0: initial reviewer proposal; makes useful-candidate retention explicit, prefers narrow revision over rejection when a grounded core remains, preserves supported procedures, and forbids replacing removed speculation with a new unsupported action. 2026-07-27: compressed repeated grounding and edge-case guidance without changing the decision or output contract."
 variables:
   - agent_context_prompt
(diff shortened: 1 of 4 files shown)


guangyu-reflexio:
Both findings addressed.

**1. Check-7 precedence listed after `revise`** — correct, and already fixed in `98d37ede` (pushed before this review landed, so it wasn't in the reviewed diff). The check-7 rule now sits at position 1, ahead of `accept` and `revise`. Worth noting the failure was observed empirically before the fix: one measurement returned `absence_inference` attached to a `revise`, which I initially misattributed to the model ignoring an instruction. It was following the stated order exactly.

**2. `absence_inference` not enforced as reject-only** — correct, and a genuine gap I had missed. Fixed in `00872b80` with an after-validator on `CandidateReviewDecision`, plus a test covering both invalid pairings, the valid reject pairing, and that other reason codes still reach `revise` unaffected.

The second finding is a sharper version of what this PR is about. I fixed the ordering in the prompt and assumed prose would carry the invariant — the same mistake the change itself targets. A rule stated only in wording gets followed most of the time; making it structural is what turns "usually" into "always".

`706 passed, 5 skipped`.

PR #434, opened 2026-08-07
test: gate live-provider tests on the credential they actually need

yilu331:
## Problem

Two skip gates keyed on `OPENAI_API_KEY`/`ANTHROPIC_API_KEY`, so **72 tests skipped silently** on a MiniMax-only machine. A sweep of every skip gate in the suite found these are the only two provider-key gates; `test_litellm_strict_structured_compat.py` already treats MiniMax as a first-class `ProviderCase` and is untouched.

Underneath that sat a second problem: neither file could reach a real provider from where it lived. `llm_mock._is_e2e_test_run` exempts only paths containing `e2e_tests` from the session-wide `litellm.completion` patch, so "live" tests placed anywhere else assert against canned mock text regardless of markers or keys.

## Changes

**1. `real_generation_provider()`** (`test_support/llm_credentials.py`), plus a public `provider_env_vars()` accessor so it doesn't reach into `model_defaults._ENV_TO_PROVIDER`. The provider-agnostic counterpart to `real_provider_key`, for tests that resolve their model via `resolve_model_name(ModelRole.GENERATION)` rather than naming one. 5 unit tests added.

**2. `test_reviewer_manifest_regression.py`** → `tests/e2e_tests/test_reviewer_manifest_regression_real_llm.py`, regated on `real_generation_provider({"openai", "anthropic", "minimax"})`. Restricted to a named set rather than every generation-capable provider, so a weaker model behind some other key can't turn silent skips into false regressions.

It passed **12/12 against `minimax/MiniMax-M3`** once able to run — `elapsed_seconds=9.889`, `cost: $0.000749` per call. From the old path, forcing the gate open produced 12 schema-parse failures in 0.88s against the mock.

This also exposed two latent bugs, invisible while the file could only skip: `request_context` was a `SimpleNamespace` where `RequestContext` was required, and the rendered `PlaybookPromptContext` was passed as `agent_context` — a slot `decide()` builds itself and fills from `configurator.get_agent_context()` in production.

**3. `test_litellm_client.py`** gated all 60 tests at module level, but most build a client from a literal config and assert how it resolved. The credential gate and `integration` marker moved off the module; **36 tests now run in the default unit tier** instead of being skipped everywhere including CI.

**4. The live half of that file** (24 tests) moved to `tests/e2e_tests/test_litellm_client_real_llm.py` for the same mock-bypass reason, with an autouse guard asserting `MOCK_LLM_RESPONSE` is unset. Its provider-specific gate stays: this half *is* the OpenAI/Anthropic wiring test (`test_switch_from_openai_to_claude`, Azure endpoint resolution, `APIKeyConfig(openai=..., anthropic=...)` routing), so converting it to MiniMax would delete the coverage.

## Verification

- `4365 passed, 9 skipped` — full OSS unit tier
- `36 passed` in the default tier for `test_litellm_client.py`, against a 60-skipped baseline
- `12 passed in 110.46s` against real MiniMax for the manifest regression
- Both e2e files skip cleanly with every provider key blanked
- Mock bypass confirmed: with a bogus-but-valid-looking `OPENAI_API_KEY`, the moved client test reaches OpenAI and fails `AuthenticationError: Incorrect API key provided` — it hit the network, not the mock
- ruff + pyright clean

**Not verified:** the 24 moved client tests' assertions against real OpenAI/Anthropic — no keys for either provider on the machine this was developed on. The move is proven correct (network reached); the assertions themselves have never run and should be exercised by someone holding those keys.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end coverage for real AI providers, including generation, chat, structured output, embeddings, multimodal requests, model switching, and API-key overrides.
  * Improved automatic detection of available provider credentials and support for provider-specific test execution.
  * Expanded safeguards to ensure real-provider tests run without mocked AI responses.
  * Converted core client tests to reliable offline unit tests while retaining configuration and edge-case coverage.
  * Added regression coverage for mock restoration and reviewer behavior.

### DIFF
--- a/reflexio/server/llm/model_defaults.py
+++ b/reflexio/server/llm/model_defaults.py
@@ -87,6 +87,21 @@ _API_KEY_CONFIG_FIELDS: dict[str, str] = {
 }
 
 
+def provider_env_vars() -> dict[str, str]:
+    """Return the env-var-to-provider mapping used for credential detection.
+
+    A read-only view for callers that need to ask about a *specific* credential
+    rather than "what is available" — notably the test suites' live-provider
+    skip gates, which must distinguish a real key from the placeholder that
+    ``ensure_provider_credential`` pins. Providers reached through a local CLI
+    rather than an API key (``claude-code``) have no entry here.
+
+    Returns:
+        dict[str, str]: Copy of the env var name to provider key mapping.
+    """
+    return dict(_ENV_TO_PROVIDER)
+
+
 def detect_available_providers(
     api_key_config: APIKeyConfig | None = None,
 ) -> list[str]:
--- a/reflexio/test_support/llm_credentials.py
+++ b/reflexio/test_support/llm_credentials.py
@@ -83,9 +83,15 @@ def real_generation_provider(
     machine whose only credential is, say, MiniMax -- even though model
     resolution would have picked MiniMax and the test would have run.
 
-    Selection follows the same priority order as
-    :func:`detect_available_providers`, so the provider named here is the one
-    ``resolve_model_name`` will pick.
+    Selection walks :func:`detect_available_providers` in its priority order
+    and returns the first eligible provider holding a real key. That matches
+    what ``resolve_model_name`` picks whenever every higher-priority detected
+    provider also holds a real key -- which the credential floor cannot
+    violate, since :func:`ensure_provider_credential` only ever pins
+    ``OPENAI_API_KEY`` and ``openai`` sits last in the priority order. An
+    environment that hand-sets some *other* provider's variable to the
+    placeholder string can still diverge; the caller sees a skip reason rather
+    than a wrong model in that case.
 
     Args:
         allowed (Collection[str] | None): Restrict to these provider keys, e.g.
--- a/reflexio/test_support/llm_credentials.py
+++ b/reflexio/test_support/llm_credentials.py
@@ -17,10 +17,12 @@ which is why the same suite can be green locally and ~1,250 errors deep in CI.
 from __future__ import annotations
 
 import os
+from collections.abc import Collection
 
 from reflexio.server.llm.model_defaults import (
     GENERATION_CAPABLE_PROVIDERS,
     detect_available_providers,
+    provider_env_vars,
 )
 
 # Matches the string already used by the `storage` fixture in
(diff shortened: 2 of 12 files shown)


yilu331:
Pushed `f927db2`, which corrects the guard added in `4ee26f7`. It did not work.

`pytest -m e2e` — the invocation the testing conventions document — leaves the session mock **on**, because `_is_e2e_test_run` matches on the invocation path string and `-m e2e` carries no path. `configure_llm_mock` then sets `MOCK_LLM_RESPONSE=true`, and `tests/e2e_tests/conftest.py:46-47` deletes it again for exactly the `requires_credentials` tests, on the assumption the patch is already off. Flag gone, patch live: the env-var guard passes and the test asserts against canned mock text.

It surfaced as a wrong comparison rather than an error — the LiteLLM smoke test received the profile-extraction payload:

```
assert '4' in '```json\n{"add": [{"content": "prefers dark mode", ...}]}\n```'
```

`f927db2` adds `litellm_is_patched()` / `assert_litellm_unpatched()` to `llm_mock` and asks the patcher directly, which no env-var edit can erase. Same swap applied to the two pre-existing guards with the identical flaw (`test_knowledge_gap_real_llm` asserted, `test_resumable_extraction_e2e` skipped).

Verified both directions:
- under `-m e2e` → fails with `litellm.completion is patched` instead of a bogus comparison
- under `pytest tests/e2e_tests/...` → still reaches OpenAI (`AuthenticationError` on a deliberately invalid key)

`4365 passed, 9 skipped` on the OSS unit tier; ruff and pyright clean.

This makes the failure honest but does not make `-m e2e` work. The underlying design — the exemption being a property of the command line rather than of the test — is filed as #435, since fixing it changes mock behavior for every suite.

yilu331:
Two more commits.

**`5e23d38` — addressed the review.**

*`test_resumable_extraction_e2e.py`* — valid, and my own doing. That test drives a running backend over HTTP ("requires live provider credentials in that backend process"), so the pytest process's patch state says nothing about what the server does. My edit made it actively worse: under `-m e2e` the in-process patch is on, so the guard would have **skipped a test that would otherwise run correctly**. Reverted — guarding an out-of-process test needs a server-side signal, which is out of scope here.

*`real_generation_provider` docstring* — softened rather than gated. The reported failure mode (placeholder `OPENAI_API_KEY` outranking a real MiniMax key) cannot occur: `ensure_provider_credential` only ever pins `OPENAI_API_KEY`, and `openai` sits **last** in `_PROVIDER_PRIORITY`, so the floor can never outrank a real key. Divergence needs a hand-set placeholder on some *other* provider. The suggested stricter gate also checks before the `allowed` filter, so it would return `None` for providers outside the allowlist and cause spurious skips. The docstring did overclaim, so that is what I fixed.

**`346ff03` — closes #435**, folded in here because it is the other half of the same problem.

The guards in this PR make the failure honest but do not make `-m e2e` work. `_is_e2e_test_run` scans `config.args` for `"e2e_tests"`, so the exemption was a property of the command line; the e2e conftest compensated by deleting `MOCK_LLM_RESPONSE`, which removed the evidence rather than the patch.

`unpatched_litellm()` lets the conftest **stop the session patcher** for `requires_credentials` items instead. The exemption now follows the marker, so a test behaves the same way under every invocation:

```
pytest open_source/reflexio/tests -m e2e -k test_generate_response_simple
  before: assert '4' in '```json\n{"add": [{"content": "prefers dark mode"...
  after:  AuthenticationError: Incorrect API key provided
```

`assert_litellm_unpatched()` stays as a backstop for a live test reached some other way — outside `tests/e2e_tests/`, or missing the marker the fixture keys on.

Unit tier `4369 passed, 9 skipped`; standard e2e tier unchanged. ruff + pyright clean.

yilu331:
@coderabbitai review

yilu331:
`100ff08` — good catch, and a regression I introduced in `346ff03`.

`unpatched_litellm` returned early when `_litellm_patcher` was `None`, **before** clearing the flag. That is exactly the normal `pytest tests/e2e_tests/` case: `configure_llm_mock` never patches there, and `patched_litellm` deliberately leaves `MOCK_LLM_RESPONSE` set on exit ("tests that read it should not depend on its lifetime being tied to a single context block"), so a deterministic e2e test hands the flag to the next live one. `litellm.completion` is real, `assert_litellm_unpatched()` passes, and every service branching on the flag still takes its mock path — the precise silent degradation this PR exists to remove. The conftest I replaced *did* clear it via `monkeypatch.delenv`; my rewrite dropped that.

Now cleared unconditionally, with the prior state restored exactly — including removing the flag when none was set, rather than leaving one behind.

Two tests added. The stale-flag one is non-vacuous: against the early-return form it fails with

```
assert 'true' is None
```

Unit tier `4371 passed, 9 skipped`; ruff + pyright clean.

yilu331:
@coderabbitai review

PR #436, opened 2026-08-07
fix(playbook): don't abort a rerun aggregation when mock mode has no centroid

yilu331:
## Problem

4 e2e tests fail on `main`:

```
RuntimeError: rerun agent playbook has no centroid embedding
aggregator.py:1773
```

- `test_playbook_workflows.py` — 3 tests
- `test_openclaw_integration.py` — 1 test

They pass at `b0e0755` and fail from `c24e1e7` onward, so this is a regression, not an environment quirk.

## Root cause

Two changes that were each fine alone.

**`159d8ab` (#410)** added the invariant: on the rerun path, a saved agent playbook must carry an embedding to persist as a cluster centroid. Safe at the time — local embeddings were computed **in-process**, so `saved_fb.embedding` was always populated and the raise was unreachable outside a genuine bug.

**`c24e1e7` (#425)** removed in-process local inference ("service-only inference boundary"). `embedding_provider_mode` now returns `local_service` for any local model with no env vars set, so embedding goes over HTTP to `127.0.0.1:8072`.

With the embedder unreachable, `SQLiteStorage` deliberately degrades:

```
Embedding unavailable for document text; continuing without vector
```

…and saves the playbook with `embedding=None`. The invariant then fires and rolls back the entire aggregation. An assertion written to catch a *programming error* now trips on an *infrastructure state*.

This is not narrow: `reflexio/lib/_generation.py:73` sets `rerun=True` for every `run_playbook_aggregation()` call, so the branch is the normal path, not an edge case.

## Why it wasn't caught

- This repo runs no CI workflows.
- Enterprise `ci-fast.yml:104` runs `--ignore=tests/e2e_tests/`.
- The e2e tier only runs in `release.yml`.
- The invariant had **no test coverage at all** — `grep 'no centroid embedding' tests/` returns nothing.

## Fix

Mock mode is the case that has no centroid *by construction*: it clusters by trigger rather than by vector (the `MOCK_LLM_RESPONSE` branch in `get_clusters`), so a centroid was never meaningful there. Skip the cluster bookkeeping instead of aborting.

Every other caller still raises — a centroid-less cluster row would silently break the incremental re-aggregation that table exists to feed. Production behaviour outside mock mode is unchanged.

Adds the three cases the invariant never had:
- the raise still fires outside mock mode
- mock mode reaches the save, then skips the centroid write
- the happy path still records the cluster

Verified the new tests fail against the pre-fix aggregator with the exact `RuntimeError`.

## Verification

- OSS e2e tier: **47 passed, 87 skipped** (was 4 failed / 43 passed)
- OSS unit tier: **4368 passed, 9 skipped**
- ruff + pyright clean

## Worth a second opinion

The mock-mode carve-out fixes the tests, but the underlying mismatch is broader: storage treats a missing embedding as *degrade and continue*, while the aggregator treats it as *fatal*. In any deployment where the embedding service is unreachable, `run_playbook_aggregation()` now hard-fails and rolls back rather than degrading. That may well be intended — failing loudly beats silently writing a useless centroid — but it changed behaviour without discussion when #425 landed, so the owners of #410/#425 should confirm which semantics they want.

### DIFF
--- a/reflexio/server/services/playbook/components/aggregator.py
+++ b/reflexio/server/services/playbook/components/aggregator.py
@@ -1769,26 +1769,37 @@ class PlaybookAggregator:
                         playbook_aggregator_request.rerun
                         and self.aggregation_claim is not None
                     ):
-                        if not saved_fb.embedding:
+                        # Mock mode clusters by trigger rather than by vector
+                        # (see the MOCK_LLM_RESPONSE branch in
+                        # ``get_clusters``), so no centroid exists
+                        # to persist and cluster bookkeeping is skipped. Every
+                        # other caller still aborts on a missing embedding: a
+                        # centroid-less cluster row would silently break the
+                        # incremental re-aggregation this table exists to feed.
+                        if (
+                            not saved_fb.embedding
+                            and os.getenv("MOCK_LLM_RESPONSE", "").lower() != "true"
+                        ):
                             raise RuntimeError(
                                 "rerun agent playbook has no centroid embedding"
                             )
-                        cluster_id = self._stable_aggregation_cluster_id(fp_key)
-                        self.storage.create_playbook_aggregation_cluster(  # type: ignore[attr-defined]
-                            cluster_id=cluster_id,
-                            agent_version=self.agent_version,
-                            agent_playbook_id=saved_fb.agent_playbook_id,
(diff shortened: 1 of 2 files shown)


yilu331:
@coderabbitai review

Issue #435, opened 2026-08-07
Mock exemption keys off the pytest command line, not the test — `-m e2e` silently mocks live-provider tests


yilu331:
## Summary

`llm_mock._is_e2e_test_run` decides whether to patch `litellm.completion` by scanning `config.args` for the substring `"e2e_tests"`. Nothing about a test's location, markers, or credentials enters the decision. The same test is live or mocked depending on how the command was typed.

```python
# reflexio/test_support/llm_mock.py
def _is_e2e_test_run(config: Any) -> bool:
    args = config.args if hasattr(config, "args") else []
    for arg in args:
        if "e2e_tests" in str(arg):
            return True
```

| Invocation | Mock | Where it comes from |
|---|---|---|
| `pytest tests/e2e_tests/` | off | `release.yml:78` — CI is fine |
| `pytest -m e2e` | **on** | the documented way to run the e2e tier |

`pytest -m e2e` is what the testing conventions tell people to use for the e2e tier. Under it, every live-provider test runs against the mock.

## Why it stays hidden

These tests don't error — they assert against canned text from whatever unrelated prompt the mock's heuristics matched. Running a LiteLLM smoke test this way:

```
assert '4' in '```json\n{"add": [{"content": "prefers dark mode", "time_to_live": "one_month"}], ...}\n```'
Token usage - model: gpt-5.4-mini, input: <MagicMock ...>
elapsed_seconds=0.006
```

A client smoke test received the profile-extraction payload.

The obvious tripwire doesn't work either. `configure_llm_mock` sets `MOCK_LLM_RESPONSE=true` when it patches, but `tests/e2e_tests/conftest.py:46-47` deletes that variable for exactly the `requires_credentials` tests, assuming the patch is already off:

```python
if request.node.get_closest_marker("requires_credentials"):
    monkeypatch.delenv("MOCK_LLM_RESPONSE", raising=False)
```

So under `-m e2e` the patch is live and the flag is gone: any env-var guard passes.

PR #434 works around this by asking the patcher directly (`assert_litellm_unpatched()`), which is unforgeable. That turns a silent wrong-assertion into a loud failure, but it does not make the invocation correct — the tests still cannot run under `-m e2e`.

## Proposal

Make the exemption a property of the test rather than the command line. Options, roughly in order of preference:

[...]

PR #440, opened 2026-08-07
test: close the two live-provider guard gaps from #435

yilu331:
The two follow-ups recorded when #435 was closed.

## Gap 1 — two live tests had no backstop

`test_glm_fallback_real_llm.py` and `test_search_eval_real_llm.py` carried no `assert_litellm_unpatched()`.

Fixed one level up instead of four inline asserts: the e2e conftest now asserts immediately after it lifts the session patch for a `requires_credentials` test. Every live test under `tests/e2e_tests/` gets the backstop — including ones not written yet — and it additionally verifies the *lift itself* took. Anything that patched litellm at another layer, or a future change that quietly makes the suspend a no-op, now fails at fixture setup instead of grading canned text.

Verified it fires. With `unpatched_litellm`'s suspend neutered to a no-op:

```
1 error
litellm.completion is patched
```

instead of the previous silent `assert '4' in '```json\n{"add": [{"content": "prefers dark mode"...`.

## Gap 2 — an out-of-process test needs an out-of-process signal

`test_resumable_extraction_e2e` drives a separate server ("requires live provider credentials **in that backend process**"). Nothing in the pytest interpreter describes it: `litellm_is_patched()` reports on pytest's own litellm, and a local `MOCK_LLM_RESPONSE` is read by the wrong process. Both pass while the backend serves canned payloads.

`/healthz` now reports the worker's own `MOCK_LLM_RESPONSE`, and the test skips on it.

That field earns its place beyond this test: a worker answering from the canned mock is a misconfiguration anywhere real answers are expected, and it is **not otherwise observable from outside the process** — a caller gets plausible-looking text either way. `/healthz` is already the diagnostics endpoint and is installed unconditionally (`api.py:680`).

A backend too old to report the field leaves the answer unknown; the run then proceeds exactly as it does today rather than skipping on a missing key.

## Note on scope

This adds a field to a production response (`/healthz`), which is more than a test-only change. It is additive, reads one env var, and `/healthz` appears in no public docs or `docs_for_coding_agent` — only historical plan files. Happy to drop it and leave gap 2 documented-but-unfixed if you'd rather not grow that payload; there is no purely test-side way to observe another process's mock state.

## Verification

| | |
|---|---|
| OSS unit tier | `4387 passed, 9 skipped` |
| OSS e2e tier | `47 passed, 87 skipped` |
| `-m e2e` live test | still reaches the provider (`AuthenticationError` on a deliberately invalid key) |
| backstop | fires on a neutered suspend (shown above) |
| ruff + pyright | clean |

Two new `/healthz` tests cover the field, including that only an explicit `"true"` counts.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end reliability by validating credentials and provider configuration before real-LLM scenarios.
  * Corrected workflow checks for playbook generation, resumable extraction, concurrent extraction, profile changes, and interaction search.
  * Ensured extraction runs consistently once per publish operation.
  * Added safeguards confirming tests run without unintended mocked responses.
  * Expanded coverage for mock-mode detection, real-provider execution, fallback behavior, and knowledge-gap scenarios.

### DIFF
--- a/reflexio/server/api_endpoints/health_api.py
+++ b/reflexio/server/api_endpoints/health_api.py
@@ -73,23 +73,12 @@ def install(app: FastAPI) -> None:
 
     @app.get("/healthz")
     def healthz() -> dict[str, Any]:
-        """Return per-worker process metrics for liveness/observability.
-
-        ``mock_llm_response`` reports whether this worker answers LLM calls
-        from the canned test mock instead of a provider. It is a
-        misconfiguration anywhere real answers are expected, and it is not
-        otherwise observable from outside the process -- a caller sees
-        plausible-looking text either way. Out-of-process tests that assert
-        real model behavior read it to skip rather than assert against canned
-        payloads.
-        """
+        """Return per-worker process metrics for liveness/observability."""
         return {
             "pid": os.getpid(),
             "uptime_sec": time.monotonic() - _STARTED_AT,
             "request_count": _REQUEST_COUNT,
             "rss_mb": _read_rss_mb(),
-            "mock_llm_response": os.getenv("MOCK_LLM_RESPONSE", "").strip().lower()
-            == "true",
         }
 
     @app.get("/healthz/eval")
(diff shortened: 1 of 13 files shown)


yilu331:
@coderabbitai review

Issue #428, opened 2026-08-07
reason_code vocabulary leaks a 13th value through an untyped public field


guangyu-reflexio:
Found during review of #426. Pre-existing.

`review_service.py` emits `reason_code="evidence_unavailable"` into `ReviewUserPlaybookResult.reason_code`, which is declared as a plain `str` in `models/api_schema/domain/entities.py` — not the `Literal` used by `CandidateReviewDecision`.

So the public vocabulary is effectively `Literal ∪ {evidence_unavailable}`, enforced nowhere and documented nowhere.

Backward-compatible in the strict sense (the field was never an enum), but:

- API consumers branching on the value silently gain unhandled cases whenever the reviewer vocabulary grows. #426 added `not_agent_decision` with no OpenAPI signal.
- The taxonomy itself is not MECE — `unsupported_evidence` is a genus and `absence_inference`, `unsupported_causality`, `unseen_artifact`, `speculative`, `internal_status` are all species of it. Measured consequence: six near-identical candidates asserting the same defect drew four different reject codes, plus a revise, plus an `accept`.
- Three codes (`speculative`, `redundant`, `internal_status`) have no numbered check defining when they apply, so the model picks among overlapping labels with no tie-breaker.

Worth deciding whether the public field should be typed, and separately whether the codes need a strict precedence order (most specific species first, `unsupported_evidence` as the catch-all genus last). The prompt already does this for checks 7–8 — "Checks 7-8 only" — and that is the one place labelling is consistent.

guangyu-reflexio:
Vocabulary half fixed in #439 (+ enterprise #967 for the public docs).

`ReviewUserPlaybookResult.reason_code` is now a 13-value `Literal` — the reviewer's 12 plus `evidence_unavailable` — defined once in `models/api_schema/domain/enums.py` and imported by both the reviewer and the public entity, so the enforced schema and the advertised schema cannot drift. The JSON schema advertises one enum instead of an untyped string, and the docs enumerate it.

**Leaving this open for the MECE half**, which is the harder and more interesting part:

- `unsupported_evidence` is a **genus**; `absence_inference`, `unsupported_causality`, `unseen_artifact`, `speculative` and `internal_status` are all species of it. When a candidate is an absence-inference, *both* labels are literally correct — the model faces a genuine tie, not a comprehension failure.
- `speculative`, `redundant` and `internal_status` appear in the enum and in decision rule 4 but have **no numbered check** defining when they apply.
[...]

PR #446, opened 2026-08-14
feat(reviewer): ask what an entry is ABOUT before asking if it is supported

guangyu-reflexio:
Addresses the leak behind #428. **Reverses my own earlier conclusion** — see the [correction](https://github.com/ReflexioAI/reflexio/issues/428#issuecomment-5287906474).

## The problem is recognition, not precedence

Decision rule 1 already said *"always `reject` a check-7 or check-8 failure"*. It fired inconsistently anyway — because the reviewer was not classifying these entries as check-8 failures at all, so the ordering had nothing to act on.

Its own `reason` text shows the split. On a correct reject:

> "The entry's subject is another system's FIRST_LECTURE_READY notification failure output, not a decision the agent itself makes…"

On one that leaked:

> "**Core lesson (rephrase rather than echo opaque notification) is grounded** in three failure events. Removed clauses about identifying an 'affected user'…"

It goes straight down the evidence axis and treats *"rephrase rather than echo that output"* as a core lesson to **preserve** — precisely the subject check 8 disqualifies. Reordering labels cannot fix a step that never runs.

## The change

Name the subject **first**, before evidence sufficiency is considered at all, and say plainly that trimming unsupported clauses does not rescue such an entry ("not a core to preserve, it is the same subject in gentler words").

## Measured

Paired over frozen candidate pools — one first-pass pool per window, every arm replayed over the same pool — **24 runs, two tenants**:

| window | base | **gate** | narrow |
|---|---|---|---|
| **712** known-bad | 8/20 leaked | **1/20** `p=0.039` | 5/20 `p=0.508` |
| **580** healthy, same tenant | 48/58 | **53/58** (+5) | 50/58 |
| **org 43 pb55** cross-tenant | 59/72 | 53/72 `p=0.238` | 60/72 `p=1.000` |
| **org 43 pb29** cross-tenant | 42/48 | 43/48 `p=1.000` | 38/48 `p=0.388` |

The leak falls **8/20 → 1/20** while healthy-window retention **rises** 48/58 → 53/58. Cross-tenant differences are within noise in both directions.

## The explanatory clauses are load-bearing

The narrow variant — same rule, explanation removed — leaks **5/20** and is not significant. So `required_invariants` now anchors those phrases, and the guard is **mutation-checked**: replacing the rule with the narrow form fails it.

## Why an earlier evaluation said the opposite

I previously measured this rule as destroying 4 healthy entries and rejected it. That evaluation was invalid — the harness passed a `PlaybookPromptContext` into `decide()`'s `str`-typed `agent_context` slot, so every prompt carried a duplicated evidence blob, and the inflation penalised the longest arms most. Fixed in enterprise #968; caught by pyright, which I had initially dismissed as environment noise.

## Caveats worth carrying

- `p=0.039` is one significant result on one window shape (tenants whose published agent turns are largely machine-emitted templates). It is not a general claim about reviewer quality.
- Single runs are useless here — the 712 baseline swung 0/3/3/4/4. Everything above is pooled over two run-pairs per arm.

## Test plan

```
tests/server/services/playbook/ + tests/server/prompt/ + prompt mapping
  -> 648 passed, 5 skipped
```
Budget 1260 → 1360. Version pins bumped to v1.3.0; v1.2.0 deactivated.

## Summary by CodeRabbit

* **New Features**
  * Added an updated playbook candidate review workflow with clearer decision ownership, evidence requirements, and structured review outcomes.
  * Introduced stricter handling for unsupported absence claims and decisions based on other systems’ outputs.

* **Bug Fixes**
  * Improved review guidance to prevent unnecessary rewording and ensure consistent candidate decisions.

* **Tests**
  * Updated validation to use the new review workflow version and confirm its expanded guidance.


guangyu-reflexio:
Thanks — took the second one, declining the first, with reasoning.

**Positional assertion: done.** You were right that independent substring checks would stay green through a reordering regression, which is the one regression that matters here. Now asserts the gate precedes the accept/revise branches, mutation-checked.

**Moving the gate above the evidence checklist: not in this PR.** The concern is fair on a plain reading — the document does still list `For each candidate, verify` (evidence) at line 25 and the subject gate at line 69. But two things:

1. **This exact wording and layout is what was measured.** 24 paired runs over frozen candidate pools: leak 8/20 → 1/20 (`p=0.039`), healthy-window retention 48/58 → 53/58. Restructuring the document changes the artifact, and this check is demonstrably dose-dependent — a variant with only the explanatory clauses removed leaks 5/20 (`p=0.508`). I would be shipping an unmeasured prompt on a theoretical argument.

2. **The evidence says the ordering already works as written.** `not_agent_decision` fired 19/20 on the target window. The failure mode you describe — reviewer takes the evidence path first — is exactly what this PR fixes, and the measurement shows it fixed. The checklist enumerates what to *verify*; the decision procedure states the *order*, and the gate is rule 1 of it.

If you would still like the document restructured, that is a reasonable follow-up — but it needs its own measurement run rather than riding along here. I would rather ship the version with numbers behind it than the tidier one without.

PR #475, opened 2026-08-31
feat: contract the replay vocabulary and fix the allowlist removal path

guangyu-reflexio:
The open-source half of Phase 7 (the §21 clean break): contracts the replay literals out of the public type surface and fixes the SQLite allowlist rebuild so a *removal* actually takes effect.

Stacked on ReflexioAI/reflexio#474 (Phase 6). Review only the commits above that base.

## The defect this fixes, which is the interesting part

Both SQLite allowlist rebuilds trigger on a **positive** substring predicate:

```python
if all(check in table_sql for check in required_checks) and (
    "'offline_tuner_open_world'" in table_sql
):
    return
```

That is correct for *adding* a literal and silently wrong for *removing* one. Drop a literal from the target tuple and an existing database — whose `table_sql` still contains every remaining required string — satisfies `all(...)` and returns early. **The rebuild never runs and the old permissive CHECK survives forever.**

A migration that reports success while changing nothing. The removal direction needs a negative predicate: rebuild if any *retired* literal is still present.

Proven by contrast rather than asserted:

| `_base.py` triggers | result |
| --- | --- |
| negative predicate (shipped) | 4 passed |
| reverted to positive-only | 2 failed — `assert 'offline_tuner_replay' not in 'CREATE TABL…'` |

## Remediation, because the rebuild is a table copy

A row carrying a retired value fails the copy, so the copy remediates: `offline_tuner_replay` → `offline_tuner_legacy` (accurate — those jobs *were* offline-tuner jobs, and that literal stays admissible), retired `stage`/`terminal_outcome` values nulled (both columns nullable), and `replay_manifest` artifact rows deleted (the column is NOT NULL and part of a UNIQUE key, so there is nothing to relabel it to).

This is local-development SQLite, not tenant Postgres. No production row is touched. Verified end to end: a legacy database carrying a replay job and a `replay_manifest` artifact migrates rather than failing, the job returns relabelled with `stage`/`terminal_outcome` NULL, the `candidate` artifact survives, and the rebuild is idempotent.

## Type surface

`optimizer_kind`, `stage`, `terminal_outcome` and `artifact_kind` unions lose their replay members. Four kinds survive.

**Backward compatibility, stated precisely:** this is a **read** break, not a write break. An older image reads a retired-literal row fine — its own `Literal` still admits it. The *new* image raises `ValidationError` at model construction. Tenant Postgres still stores such rows until the enterprise contract migration lands, so during that bake a new task reading an old replay row raises. That is the clean break by design; the enterprise side removed the path that would act on such a row.

## Scope

The OSS package remains importable and self-consistent standing alone — full non-e2e tier: **6,379 passed, 13 skipped, 0 failed**, run from the package's own venv.

## Summary by CodeRabbit

* **Breaking Changes**
  * Retired replay-specific optimizer types, job stages, terminal outcomes, and artifacts are no longer supported.
  * Publication claims and legacy publication requests now support GEPA only; open-world optimization remains supported for decision proofs.

* **Bug Fixes**
  * Database constraint validation now rebuilds incomplete schemas to preserve support for all valid optimizer values and stages.

* **Tests**
  * Expanded coverage for outcome reachability, publication validation, migrations, storage compatibility, and rejected replay transitions.

### DIFF
--- a/reflexio/models/api_schema/domain/entities.py
+++ b/reflexio/models/api_schema/domain/entities.py
@@ -436,6 +436,38 @@ OptimizationJobStage = Literal[
     "failed",
 ]
 
+# Phase 7 removed the four members whose NAMES contain 'replay'. Seven more were
+# reachable only through that same replay arm and are now equally dead: no
+# writer, no reader, no tenant routine that can set them. They are RETAINED
+# rather than removed, the way 'offline_tuner_legacy' is retained above -- and
+# for the same two reasons. Removing them narrows the tenant CHECK a second
+# time, and a CHECK narrowing is a one-way door this branch has already spent
+# once; and historical rows written before the replay retirement still carry
+# them, so a narrowed CHECK would abort the validating migration on the first
+# organization holding one.
+#
+# Read a member of this set as "an outcome an earlier era could record", never
+# as a state the current tuner can reach. Pinned by
+# tests/models/test_terminal_outcome_reachability.py so it cannot silently grow:
+# a new member added here is a new outcome somebody must show is WRITABLE.
+#
+# 'deployment_unsupported' is the trap. The same spelling is also
+# OfflineTunerUnavailableReason -- a config-enablement rejection code in
+# reflexio_ext capability_status.py, with ~40 live references. Those are a
+# different vocabulary on a different type, so "grep says it is used" does not
+# make this member reachable.
+RETAINED_UNREACHABLE_TERMINAL_OUTCOMES: frozenset[str] = frozenset(
+    {
+        "insufficient_negative_evidence",
+        "insufficient_positive_evidence",
+        "insufficient_coverage",
+        "deployment_unsupported",
+        "candidate_regressed",
+        "candidate_did_not_improve",
+        "publication_failed",
+    }
+)
+
 OptimizationTerminalOutcome = Literal[
     "applied",
     "insufficient_negative_evidence",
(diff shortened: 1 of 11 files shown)


guangyu-reflexio:
@coderabbitai review

This PR was opened against `codex/offline-tuner-open-world-phase6`, where reviews are disabled, so it was never reviewed. It has since been rebased onto `main` and retargeted. Requesting a full review of the current head.

guangyu-reflexio:
## Review disposition — the finding holds, fixed in `fe85c8bf`

It is the same class this PR exists to fix, in the direction the original fix did not cover.

The negative half of the predicate catches a **retired** literal that is still present. Nothing caught the other direction: an **admissible** literal that is **absent**. `required_checks` named only `'offline_tuner_open_world'` among the four surviving optimizer kinds, so a schema whose `optimizer_kind` CHECK omitted `'gepa'` carried no retired literal, satisfied the list, and was retained forever with a constraint that rejects legitimate gepa rows.

`'gepa'` is admissible — `OptimizerKind` (`entities.py:417-422`) and both DDL sites (`_base.py:2257`, `:3502`) all list four kinds. Three of the four went unchecked.

The artifact rebuild immediately below already enumerates its full set of eight kinds, so the optimizer rebuild was the outlier rather than the pattern. It now follows the same rule.

Proven by contrast, on a fixture that is current in every respect except the missing kind and carries no retired literal, so only the absence can trigger the rebuild:

```
pre-fix predicate  -> 2 failed, 8 passed
fixed predicate    -> 10 passed
```

Restored from a saved copy and verified by checksum (`0843a746…dd3b57f` before and after), never `git checkout --`.

Wider check: 770 passed across the OSS storage and model suites. `ruff check` clean; `ruff format` applied to the changed file.

PR #476, opened 2026-09-03
fix(playbook-optimizer): stop persisting raw exception text in decision_reason

guangyu-reflexio:
## What

`PlaybookOptimizer` wrote a raw exception string into `decision_reason`, a durable column:

```python
self.storage.update_playbook_optimization_job(
    job.job_id, status="failed", decision_reason=str(exc)
)
```

`str(exc)` on an arbitrary exception can carry customer content. That is not hypothetical — on a sibling code path a `pydantic.ValidationError` raised during playbook analysis embedded **customer transcript text**, because the model's own output is interpolated into the error message.

`decision_reason` is read back into the domain model and surfaced to operators.

## Why this shape of fix

Every other writer of that column uses a fixed, controlled string — `"assistant backend aborted one or more evaluations"`, `"best candidate did not pass commit thresholds"`, `'retired_by_replay_redesign'` — so the column is already a de facto controlled vocabulary and this line was the only outlier. A tenant Postgres trigger also compares it against literals, so the vocabulary is load-bearing rather than cosmetic.

The diagnostic is not lost: the two lines immediately above already bind `error_type=type(exc).__name__` into structured logging context and call `logger.exception`, which records the traceback where it belongs.

**The exception class name is deliberately NOT added to the column.** It buys nothing that `error_tags` does not already carry, and it would widen an operator-facing fixed vocabulary into one keyed on third-party exception types (gepa, litellm, pydantic).

## Scope

A sweep of 253 candidate expressions across both trees found 86 that reach a durable sink. **Only this one is fixed here**, on the principle that the defect is *content-bearing text written into a column not classified to hold content*. Every other DB writer targets a field whose declared purpose **is** an error string — `_agent_runs.last_error`, `operation_state.error_message`, `billing_debit_incident.detail`. Redacting those is a different design question that reaches into billing and extraction, and was not silently folded in.

The tightest sibling, recorded for follow-up: `gepa_adapter.py:257/273` writes `rationale=f"Assistant failed: {exc}"` into `playbook_optimization_evaluations.rationale`.

## Existing rows

The line has existed since 2026-05-08 (~4 months). The path is opt-in — `PlaybookOptimizerConfig.enabled=False` by default and it needs an assistant backend configured — so only orgs that deliberately enabled GEPA can hold such a row.

On enterprise Postgres these age out: `_retention_gc_retired_optimization_jobs` hard-deletes non-pending/running jobs past a cutoff, gated on `lineage_gc.enabled` (default true) with a 90-day window. So contamination is a rolling window that self-heals. **Two caveats:** OSS SQLite overrides that hook to `return 0`, so nothing purges there; and whether any org actually enabled the optimizer is an environment question that was not answered by querying anything.

## Tests

`102 passed` in `tests/server/services/playbook_optimizer/`, verified on top of `main` rather than only on the branch it was written on.

The guard asserts a **distinctive sentinel** from the exception message is absent from the persisted value — a test that merely checks "the message isn't there" passes trivially when the exception has a generic message. Mutation-verified: restoring `decision_reason=str(exc)` and the f-string variant each turn it red.

Worth noting: the optimizer's exception path had **no test coverage at all** before this change — the first run died on the shared fake request context lacking `org_id`. Four months of an untested failure path is presumably how the leak survived review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unexpected optimization errors from storing customer or provider response content in job records.
  * Failed optimization runs now show a controlled, generic failure reason.

* **Tests**
  * Added coverage to verify sensitive exception details are not persisted.
  * Added safeguards ensuring decision reasons use approved fixed text.

### DIFF
--- a/reflexio/server/services/playbook_optimizer/optimizer.py
+++ b/reflexio/server/services/playbook_optimizer/optimizer.py
@@ -249,8 +249,17 @@ class PlaybookOptimizer:
                 error_type=type(exc).__name__,
             ):
                 logger.exception("Playbook optimization failed")
+            # The exception message is deliberately not persisted here. It can
+            # carry customer content -- a pydantic ValidationError raised on a
+            # provider response renders the model's own output into its message
+            # -- and ``decision_reason`` is a durable column read back into the
+            # domain model and shown to operators. The class name, the traceback
+            # and the tags are already captured by the ``error_tags`` block
+            # above, which is where an unbounded diagnostic belongs.
             self.storage.update_playbook_optimization_job(
-                job.job_id, status="failed", decision_reason=str(exc)
+                job.job_id,
+                status="failed",
+                decision_reason="optimization run raised an unexpected error",
             )
             return "failed"
 
--- a/tests/server/services/playbook_optimizer/test_decision_reason_vocabulary_guard.py
+++ b/tests/server/services/playbook_optimizer/test_decision_reason_vocabulary_guard.py
@@ -1,0 +1,123 @@
+"""Guard: ``decision_reason`` writers must pass a fixed string, never an expression.
+
+The invariant
+-------------
+``playbook_optimization_jobs.decision_reason`` is a durable ``TEXT NOT NULL``
+column that is read straight back into ``PlaybookOptimizationJob`` and shown to
+operators, and a Postgres trigger compares it against literal values. It is a
+controlled vocabulary: every writer names a decision the optimizer made.
+
+An expression in that keyword is how customer content gets in. ``str(exc)`` on
+an arbitrary exception is the concrete case -- a pydantic ``ValidationError``
+raised on a provider response renders the model's own output into its message,
+so persisting the message persists evidence text. The class name, the traceback
+and the tags belong in the structured-logging / error-reporting path instead,
+which every failure site already uses.
+
+Why a guard rather than review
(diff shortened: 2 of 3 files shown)


guangyu-reflexio:
Superseded by #478, which merged as `8cedc3bde`.

This branch's substance is already on `main`: the `optimizer.py` change replacing `str(exc)` with the fixed `"optimization run raised an unexpected error"`, and the 123-line `test_decision_reason_vocabulary_guard.py` AST scan that keeps decision-reason writers to string literals. Both arrived via #478.

How that happened is worth recording: the commit behind this PR (`658a8ae4`) existed locally, unpushed, alongside two schema commits that were also unpushed. While unblocking an enterprise submodule gitlink that pointed at commits reachable from no remote, all of them were bundled into #478 — without noticing this one already had a PR of its own. That was my error rather than anything wrong with this branch.

Diffing the three touched files against `main` now leaves exactly one line, and it runs the wrong way:

```diff
-    key = ("org", None, "agent_playbook", 1)      # main
+    key = ("org", "agent_playbook", 1)            # this branch
```

`main` has since gained a project-scoping element in that scheduler key, so merging this branch would revert it and leave the test asserting a shape the code no longer produces. Nothing here is worth salvaging; closing rather than updating.

PR #477, opened 2026-09-03
fix(storage): resolve the SQLite file from the dataset identity

guangyu-reflexio:
## The bug

Two datasets sharing one SQLite file see each other's data. Reproduced before the fix: two
`SQLiteStorage` instances with different `org_id`s resolve to the same `db_path`, and `tenant-b`
reads `tenant-a`'s private profile content.

`SQLiteStorage.__init__` resolved `db_path` from `LOCAL_STORAGE_PATH` and never from the
caller-supplied `org_id`, so one file served every caller. Of 43 real persistent tables only 11
carry `org_id`; the other 32 — `profiles`, `requests`, `interactions`, `user_playbooks` — have no
identity column and unscoped reads.

The consequential case is **enterprise self-host with more than one org on SQLite**: the server
never passes `base_dir` and `_platform_forbids_sqlite` is False under self-host, so every org
lands on `db_path=None` and shares one `reflexio.db`.

## The fix

New `sqlite_storage/_dataset_path.py` resolves the file from the dataset identity:

- `validate_dataset_identity` — rejects unusable identities, never slugifies (slugifying maps two
  identities onto one file, which is the bug).
- `claim_or_read_identity` — first-claimer-wins under `BEGIN IMMEDIATE`, recorded in a
  `_dataset_identity` row. `BEGIN IMMEDIATE` takes a cross-process lock, needed because the
  service runs multiple workers and the existing init lock is only a `threading.Lock`.
- `resolve_sqlite_db_path` — derives `reflexio_<org_id>.db`, adopts a legacy `reflexio.db`
  in place for its first claimant, and routes every other identity to its own file.

Adopt-in-place rather than `VACUUM INTO`: copying doubles disk at upgrade with no rollback if it
dies mid-vacuum, and orphans the documented path.

## Verification

- Reproduced red, then green: distinct files, 0 rows leaked.
- Full OSS suite **1379 passed**; ruff clean; pyright 0 errors.

## Notes

- Explicit `db_path` is untouched — the regression guard for that stays unchanged, because 10
  files construct different `org_id`s against one explicit path.
- Known residual, documented not fixed: `clear-all` deletes the whole storage root as a
  directory, so one consumer's reset still destroys other identities' files in that root.
  Pre-existing and unchanged in blast radius; narrowing it is the natural follow-up.

## Summary by CodeRabbit

* **New Features**
  * SQLite storage now uses organization-specific database files when no explicit path is provided.
  * Added safeguards to validate database identity and prevent conflicting ownership claims.
  * Legacy databases can be adopted or routed to isolated organization-specific files.
  * The database reset tool now supports selecting an organization.
* **Bug Fixes**
  * Prevented accidental sharing of SQLite data between organizations.
  * Explicit database paths continue to be honored unchanged.

### DIFF
--- a/reflexio/server/services/configurator/configurator.py
+++ b/reflexio/server/services/configurator/configurator.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Callable
-from pathlib import Path
 from typing import Any
 
 from reflexio.models.config_schema import (
@@ -15,6 +14,9 @@ from reflexio.server.services.configurator.local_file_config_storage import (
     LocalFileConfigStorage,
 )
 from reflexio.server.services.storage.sqlite_storage import SQLiteStorage
+from reflexio.server.services.storage.sqlite_storage._dataset_path import (
+    resolve_sqlite_db_path,
+)
 from reflexio.server.services.storage.storage_base import BaseStorage
 
 logger = logging.getLogger(__name__)
@@ -36,11 +38,16 @@ def _create_sqlite_storage(
         full_config.enable_document_expansion if full_config else False
     )
     # When base_dir is explicitly provided (e.g. tests with temp dirs)
-    # and no db_path is configured, use base_dir for the SQLite DB
-    # so the storage is isolated from the shared default database.
+    # and no db_path is configured, resolve the SQLite DB under base_dir so the
+    # storage is isolated from the shared default database.
+    #
+    # Resolved by identity, not by base_dir alone: two orgs sharing one base_dir
+    # would otherwise land on one file and read each other's rows -- the same defect
+    # the default path had. Mirrors the config layer beside it, which already writes
+    # ``config_<org_id>.json``.
     db_path = config.db_path
     if db_path is None and configurator.base_dir:
-        db_path = str(Path(configurator.base_dir) / "reflexio.db")
+        db_path = resolve_sqlite_db_path(configurator.base_dir, configurator.org_id)
     return SQLiteStorage(
         org_id=configurator.org_id,
         db_path=db_path,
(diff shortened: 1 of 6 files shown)


guangyu-reflexio:
Closing as superseded — this work is already on `main`, landed by #481 ("fix(storage): native read transport, governance extraction, and SQLite dataset isolation", merged 2026-09-06), whose description lists "fixing SQLite dataset isolation" as one of its 16 commits.

Verified file by file against `origin/main` rather than assuming:

| File | Status on `main` |
| --- | --- |
| `sqlite_storage/_dataset_path.py` | Present. Every function this PR adds exists there; `main` additionally has `barrier_identity_labels` (342 lines vs 255 here). |
| `sqlite_storage/_base.py` | Imports `resolve_sqlite_db_path` and calls `resolve_sqlite_db_path(LOCAL_STORAGE_PATH, org_id)` — the actual fix. |
| `configurator/configurator.py` | Byte-identical to this branch. |
| `scripts/reset_db.py` | Byte-identical to this branch, `--org` included. |
| `tests/.../test_dataset_path.py` | All 10 test cases from this branch are present; `main` has 13. |
| `tests/.../test_storage_defaults.py` | No test case here that `main` lacks. |

The only differences between this branch and `main` run the other way — `main` is ahead (`init_governance_tables` replacing `init_subject_write_barrier_table`, `migrate_extraction_stream`, the `evidence_view_incomplete` literal, `learning_jobs.project_id`). Nothing in this branch is missing upstream, which is also why it now shows as conflicting.

No code is lost by closing. The residual this PR documented but deliberately did not fix — `clear-all` deleting the whole storage root, so one consumer's reset destroys other identities' files — is worth tracking as its own issue if it still applies; it was out of scope here either way.

PR #483, opened 2026-09-09
feat(extraction): resume a run under the project that owns it

guangyu-reflexio:
## The defect

The extraction resume scheduler writes project-scoped tenant tables with no project bound. `_drain_org` builds an org-only `RequestContext` and calls `ExtractionResumeWorker(...).drain(...)`, which reaches the finalization-receipt write and raises:

```
ProjectContextError: No project bound for a write to project-scoped tenant
table '_agent_run_finalization_receipts'
```

It is **latent, not live** — it needs a paused run, and neither staging environment has one. It fails closed the first time one resumes off the scheduler. Eighteen enterprise integration tests reproduce it.

This is the same class as the playbook-aggregation scheduler, on a second background writer.

## The fix

**The scope comes from the row.** A resumed run is decoupled in time from the request that created it, so there is no request-scoped project to inherit — exactly the problem `reflexio.server.work_scope` already exists to solve, and this reuses that seam rather than inventing one. The durable-learning worker binds from `job.project_id` the same way.

`_agent_runs.project_id` is stamped at insert by the storage chokepoint, which makes the row the only carrier the attribution survives on. So `AgentRunRecord` reads it back and `_scope_for(run)` turns it into a `WorkScope`.

**Read-side only on the record.** Nothing here writes `project_id`: the insert path is already stamped, and a second writer for one column is how a conflicting explicit value gets refused.

**The claims stay outside the scope, deliberately.** `claim_ready_agent_run` and `claim_finalization_failed_agent_run` are org-keyed SECURITY DEFINER RPCs that select on `org_id` alone — they neither consult nor need a bound project. Everything downstream of a claim writes project-scoped tables and does, so `run_once` binds once around `_process_claimed_run`, covering the resume, the finalization, and all four `update_agent_run_status` exit paths rather than being sprinkled through them.

**The body was moved, not retyped.** `_process_claimed_run` is byte-identical to the old `run_once` tail, extracted by script with the boundaries asserted — hand-copying eighty lines of retry/finalization branching is how a subtle behaviour change gets in.

## Runs with no project

A run whose `project_id` is NULL predates the project write path. The enterprise provider **refuses** it rather than defaulting to the org's default project, and the worker loop escalates and moves on. Those rows are fixed as *data*: the existing corpus-derived backfill already covers `_agent_runs`, `_agent_run_finalization_receipts` and `_pending_tool_calls` — verified, not assumed, so no new backfill was written.

In OSS this is all inert: `bind_work_scope` is a no-op with no provider registered, and a bare install has one org and no projects.

## Compatibility

`project_id` is appended at the **end** of the `AgentRunRecord` field list, not beside the other required fields, so positional construction is unaffected.

## Verification

- 100 passed across `tests/server/services/extraction` and `tests/server/services/durable_learning`.
- Enterprise receipt suite: **18 failed → 29 passed** (that half ships in the enterprise PR).
- ruff clean, pyright 0.
- Rebased onto current `main` and re-verified after the rebase.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when resuming interrupted runs and retrying finalization.
  - Ensured run processing and status updates consistently use the associated project context.
  - Improved handling of older or standalone run records without project information.

### DIFF
--- a/reflexio/server/services/extraction/resume_worker.py
+++ b/reflexio/server/services/extraction/resume_worker.py
@@ -84,6 +84,7 @@ from reflexio.server.services.storage.storage_base import (
 from reflexio.server.services.tagging.tagging_scheduler import schedule_tagging
 from reflexio.server.site_var.site_var_manager import SiteVarManager
 from reflexio.server.usage_metrics import UsageEventDeliveryStatus
+from reflexio.server.work_scope import WorkScope, bind_work_scope
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


guangyu-reflexio:
Verified against the schema rather than dismissed on reasoning — the finding rests on "even when the stored row has a project ID", and on SQLite no such row exists.

`_agent_runs` is created at `sqlite_storage/_base.py:3750` and has **no `project_id` column**. `data.get("project_id")` would therefore always return `None`, so the suggested line is dead code that would read as though SQLite were project-aware when it is not. The only `project_ids` anywhere in the SQLite backend is on an unrelated external-connections table (`_extras.py`).

That absence is deliberate, not an oversight: OSS is single-tenant. `work_scope`'s own docstring puts it as "a bare install has one org and no projects, so an absent project is normal, never an error", and `bind_work_scope` is inert there with no provider registered.

Checked for recurrence before clearing it, since a dismissal that holds in one path often fails in a sibling: `_row_to_agent_run` has exactly one definition in this repo (the SQLite one), and the enterprise mapper — which reads a table that *does* carry the column — already hydrates it in the paired enterprise change.

Not applying. If SQLite ever gains the column, the mapper has to be updated in the same change, and this thread is the record of why it doesn't read it today.

PR #494, opened 2026-09-11
fix(publish): stop waiting for a window only new input can close

guangyu-reflexio:
## What this fixes

#491 moved automatic extraction onto durable sliding windows. The engine is right and the design is documented — *"A partial window waits for more input or an eligible `force_extraction` request."* What did not move with it is the **caller-side wait**.

Both publish waiters still poll `extraction_status` for `"done"` against a hard 240s deadline. Under the new design that state cannot arrive until a full window of eligible interactions exists. With the default `window_size` of 10, **every `publish_interaction()` and every `POST /api/publish_interaction?wait_for_response=true` for a user below that threshold blocks for four minutes** and then reports `learning_reason="wait_timeout"` — which describes a stuck stream, not a healthy one waiting for input.

Measured directly (one interaction, profile extractor at stride 1):

```
t+ 0s  pending/queued              profile seq=0  playbook seq=0
t+ 2s  pending/waiting_for_window  profile seq=1  playbook seq=0   <- extraction DONE
t+ 4s .. t+26s  unchanged — playbook needs 9 more interactions
```

Extraction finished in ~2s. The caller was held another ~238s for nothing.

## The fix

`extraction_status` already computes the right terminal signal. `waiting_for_window` is emitted **only** when every still-pending cursor both lacks a window and cannot select one — precisely "nothing more can happen without new input". `coverage_stalled()` names it, and both waiters break on it, reporting `learning_status="deferred"` with `learning_reason="waiting_for_window"` so a healthy stream stays distinguishable from a real timeout.

A library publish goes **240s → 2.07s**, with the profile still extracted.

Fixed in **both** waiters — `generation_service.run` and the HTTP route — rather than the one that showed up in the traceback.

**Second, smaller fix:** `extraction_status`/`extraction_counts` are reporting reads made *after* the interactions commit, inside the publish's `try`. An unguarded raise there returned `success=False` for work already on disk, inviting a retry of a publish that happened. `_safe_coverage()` restores the invariant the deleted `_safe_count` helper carried in its docstring, degrading to omitted fields instead.

## Why the test suite could not catch this

It never finished. 23 tests in `test_profile_workflows_unit.py` each burned the full 240s wait, so a complete run was never going to land inside any CI timeout.

| | before | after |
|---|---|---|
| `tests/` (full) | never completes | **6219 passed, 2:00** |
| `os-unit-tests` lane | red | **5023 passed, 54s** |
| `test_profile_workflows_unit.py` | 23 failed (hung) | 27 passed, 8.8s |
| `test_api_routes.py` | 4 failed, 182s | 40 passed, 1.7s |

One cause runs through nearly all of the test debt: a fixture that says *"extract on every interaction"* via `stride_size` alone. Stride was the whole gate before #491; now a never-started cursor needs a **full window** first, so a window wider than the batch extracts nothing, forever.

- `test_profile_workflows_unit.py` (23) — add `window_size_override=1` to match the stride the fixture already set.
- e2e conftest (9) — `_CUSTOMER_SUPPORT_WINDOW_SIZE` was 20 against a 16-turn scenario. Lowered to the product default of 10, with an assertion in `sample_interaction_requests` so the relationship cannot silently drift back.
- `test_knowledge_gap_playbook_extraction` (1) — a deliberately short 8-turn scenario. Uses `force_extraction`, the documented way to say "extract this conversation on its own"; padding it would dilute the knowledge-gap signal the test checks.
- Stale mocks (11) — they predate the two new storage reads. A bare `MagicMock` is not subscriptable, so `status["status"]` raised `TypeError` → HTTP 500.
- Four pinned a contract #491 deliberately replaced (`BackgroundTasks`, `"queued"`, count-delta snapshots, the pre-`not_tracked` status set). Rewritten to the new contract, each with its rationale recorded in place rather than deleted quietly.

## Attribution — verified, not assumed

The e2e failures are **pre-existing**, not caused by this change. Checked by restoring `reflexio/` to unmodified `origin/main` and re-running: `test_search_profiles_end_to_end` fails with the **identical** assertion, in **241.10s** instead of ~1s. This PR makes them fail fast, then fixes them.

`test_sync_tail_timeout_keeps_durable_pending_work` is the one test whose observed behaviour this change genuinely alters: it published 1 interaction, so it now returns early and no longer reaches the timeout branch it exists to cover. Given a full window so it still tests a real timeout, with a sibling test covering the early return — rather than relabelling it to pass.

`test_no_log_call_interpolates_a_raw_request_id` caught a raw `request_id` in the new warning line. Wrapped in `sanitise_for_log`. The guard worked.

## Verification

- Full OSS suite: **6219 passed, 0 failed** (`-m "not requires_credentials"`)
- `os-unit-tests` lane command verbatim: **5023 passed**
- OpenClaw plugin lane step: **158 passed**
- Enterprise unit suite (12 consumers #1122 adapted to #491): **8204 passed**
- `ruff check` + `ruff format --check` on the CI-gated paths: clean
- `pyright`: **0 errors**
- Runtime reproduction of the defect itself, not just the tests: publish returns in 2.07s with `learning_reason="waiting_for_window"` and the profile extracted.

## Not in scope, reported instead

`durable_learning/local.py` never removes a scheduler from `_schedulers`, so each new `storage_base_dir` leaks a polling thread for the process lifetime. It shows up in the pytest stack dumps and as `Durable extraction discovery failed` log spam once temp dirs are cleaned up. Real, but not what makes the suite red — flagging rather than widening this change.

## Summary by CodeRabbit

* **New Features**
  * Publish requests now return promptly when learning is waiting for a complete input window.
  * Deferred responses include a `waiting_for_window` reason and can be monitored through the learning-status endpoint.
  * Added documentation for default window behavior and the `force_extraction` option.

* **Bug Fixes**
  * Successfully committed publishes no longer fail when post-publish learning metrics cannot be reported.
  * Learning status now accurately distinguishes completed, deferred, and untracked processing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,3 +431,14 @@ Different users can extract concurrently within `REFLEXIO_PUBLISH_LEARNING_WORKE
 A partial window waits for more input or an eligible `force_extraction` request.
 Provider fallback remains enabled. Existing extracted data is preserved, and
 interactions stored before stream admission are not automatically replayed.
+
+A caller waiting for coverage — `publish_interaction()`, or `POST
+/api/publish_interaction?wait_for_response=true` — is never held for a window only
+new input can close. It returns as soon as every eligible cursor is either covered
+or waiting for input, reporting `learning_status="deferred"` with
+`learning_reason="waiting_for_window"`. Poll `GET /api/learning_status` with the
+returned `request_id` to see coverage complete later. Note the practical
+consequence of the default `window_size` of 10: a brand-new user's first window
+needs ten eligible interactions before anything is extracted, so a single publish
+legitimately returns uncovered. Pass `force_extraction` when you need that publish
+extracted on its own.
--- a/reflexio/lib/_interactions.py
+++ b/reflexio/lib/_interactions.py
@@ -1,9 +1,13 @@
+import logging
+from typing import Any
+
 from reflexio.lib._base import (
     STORAGE_NOT_CONFIGURED_MSG,
     ReflexioBase,
     _require_storage,
 )
 from reflexio.lib._storage_labels import describe_storage
+from reflexio.models.api_schema.common import sanitise_for_log
 from reflexio.models.api_schema.retriever_schema import (
     GetInteractionsRequest,
     GetInteractionsResponse,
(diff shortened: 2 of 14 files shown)


guangyu-reflexio:
Addressed the review finding in `56ffbbed`, reproduced before fixing rather than taken on faith:

```
success=False  message='stream read unavailable'  request_id=None
interactions actually committed: 1
```

The report was right and understated it slightly — the response also carried `request_id=None`, so the caller was told their publish failed *and* left with no id to poll for the work that had landed. The waiter performs no writes, so a read failure there now stops the wait and leaves coverage unknown, which the response already models.

This was my own fix-by-line mistake: the first commit contained this class in `_interactions.py` only and missed the sibling instance one frame up. The new regression test exercises the **real** service — the existing sibling in `test_interactions_unit.py` mocks `GenerationService` out entirely and so could never reach the waiter, which is exactly why the gap survived.

Also reworked `test_sync_tail_timeout_keeps_durable_pending_work`: at a full window the 0.1s deadline started expiring during *admission* rather than during the wait, giving a flaky 504 under parallel load. It now admits one interaction against a deadline generous enough to commit, holds coverage at `queued`, and asserts the wait actually ran before timing out.

Deliberately **not** widened to the other post-commit calls in `run()` (`_schedule_post_publish_evaluations`, `_emit_publish_success_events`, `ensure_local_extraction`) — those predate #491 and are side-effect dispatch rather than reporting reads.

Full OSS suite after the change: **6220 passed, 0 failed**.

PR #498, opened 2026-09-11
fix(durable-learning): retire library-local schedulers once their handles die

guangyu-reflexio:
## The leak

`ensure_local_extraction()` starts a `DurableLearningScheduler` per `storage_base_dir` and records it in a module-level dict that nothing ever removes. It is called from `ReflexioBase.__init__`, so **every `Reflexio(...)` built against a new directory leaks a polling thread for the life of the process**, and `_contexts` grew unboundedly alongside it.

Measured directly — six handles over six directories, each dropped and `gc.collect()`ed:

```
              BEFORE                     AFTER
build 0   threads=1  schedulers=1    threads=1  schedulers=1
build 3   threads=4  schedulers=4    threads=1  schedulers=1
build 5   threads=6  schedulers=6    threads=1  schedulers=1
```

The visible symptom was a flood of `Durable extraction discovery failed` in test output — orphaned schedulers polling storage nobody holds any more. Across a full-suite run that went from a continuous flood (90+ suppressed per 5-second window) to **zero occurrences**.

## The fix

Hold contexts in a `weakref.WeakValueDictionary` and retire a directory's scheduler once no context for it is reachable.

*"A live `Reflexio` handle points at this directory"* is the only honest signal that a library-local scheduler is still needed. Reference counting gives the same answer but requires an explicit `close()` the public API doesn't have — every caller would have to remember it, and the ones who forgot would leak exactly as before. A weak registry gets that answer for free and can't be forgotten.

**Retirement runs on the caller's thread, outside `_lock`.** `ThreadedScheduler.stop()` joins the scheduler thread, and that thread's own `discover()` callback takes `_lock` — so stopping under the lock would stall until the join timed out and leave the thread alive anyway. `adopt_server_scheduler` already had this shape; the sweep follows it.

Consequence, stated plainly: at most **one** scheduler outlives its handles, until the next `ensure_local_extraction` call. Nothing is lost — durable work is persisted, and a later `Reflexio` on that directory resumes the backlog through the existing restart-recovery path.

`adopt_server_scheduler` is unchanged, and the server path is unaffected.

## What was deliberately *not* done

The `logger.exception("Durable extraction discovery failed")` in `scheduler.py` is untouched. It was the symptom, not the bug; silencing it would have made the noise go away while hiding real discovery failures. It went quiet on its own once the orphans stopped.

## Verification

| Gate | Result |
| --- | --- |
| Full OSS suite (`-m "not requires_credentials"`) | **6222 passed, 0 failed** |
| `Durable extraction discovery failed` in that run | **0** (was a continuous flood) |
| `tests/server/services/durable_learning/` | 5 passed (3 pre-existing + 2 new) |
| `tests/server/services/test_durable_window_pipeline.py` | 19 passed |
| `tests/lib/test_profile_workflows_unit.py` | 28 passed |
| `ruff check` / `format --check` | clean |
| `pyright` | 0 errors |

The before/after thread counts above were produced by reverting `local.py` to its pre-fix content, re-running the same probe, and restoring by checksum (`234a6c50…` confirmed identical afterwards) — not by `git checkout --`, which this project's rules flag as having silently destroyed work before.

One honesty note: an earlier attempt to reproduce the log spam by `rmtree`-ing the temp directory produced **zero** failures on the unfixed code, because SQLite keeps working through its open fd on macOS. The probe above reproduces the *mechanism* — a dropped handle's storage still being polled — rather than that exact trigger. The thread leak itself is reproduced and fixed directly.

## New test

`tests/server/services/durable_learning/test_local_lifecycle.py` — two tests, one per half of the invariant: a scheduler survives while a handle is alive, and is retired once none is.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lifecycle management for local learning schedulers.
  * Independent learning contexts remain isolated when sharing organization and storage settings.
  * Schedulers are retired when their contexts are no longer reachable, while reachable contexts keep schedulers running.
  * Prevented cross-directory context discovery and scheduler interference.

* **Tests**
  * Added coverage for scheduler cleanup, context isolation, and preventing thread accumulation.

### DIFF
--- a/reflexio/server/services/durable_learning/local.py
+++ b/reflexio/server/services/durable_learning/local.py
@@ -1,17 +1,100 @@
 """Library lifecycle registration for the same durable extraction scheduler."""
 
 import threading
+import weakref
 
 from reflexio.server.api_endpoints.request_context import RequestContext
 from reflexio.server.services.durable_learning.scheduler import DurableLearningScheduler
 
 _lock = threading.RLock()
-_contexts: dict[tuple[str, str | None], RequestContext] = {}
+
+# Lifecycle choice: contexts are held WEAKLY, and a directory's scheduler is
+# retired once every context for it has been collected.
+#
+# A library-local scheduler exists to serve live ``Reflexio`` handles, so "a
+# handle pointed at this directory is still reachable" is the only honest
+# signal that one is still needed. Reference counting would have said the same
+# thing but needs an explicit ``close()`` on a public API that has none today —
+# every caller would have to remember it, and the ones that forgot would leak
+# exactly as before. A weak registry gets the same answer for free and cannot
+# be forgotten.
+#
+# Retirement runs on the *caller's* thread inside ``ensure_local_extraction``
+# rather than from inside a tick: ``stop()`` joins the scheduler thread, and a
+# thread cannot join itself. Nothing is lost by waiting for the next caller —
+# durable work is persisted, and a dropped handle means nobody in this process
+# is waiting for its results; a later ``Reflexio`` on the same directory picks
+# the backlog up through the usual restart-recovery path.
+_contexts: "weakref.WeakValueDictionary[tuple[str, str | None], RequestContext]" = (
+    weakref.WeakValueDictionary()
+)
 _schedulers: dict[str | None, DurableLearningScheduler] = {}
 _server_scheduler: DurableLearningScheduler | None = None
 
 
+def _start_scheduler(directory: str | None) -> DurableLearningScheduler:
+    """Build and start the scheduler that polls one storage directory.
+
+    Both closures capture only ``directory``, never a context, so the registry
+    stays the sole owner of context lifetime.
+
+    Args:
+        directory (str | None): Storage base directory the scheduler serves.
+
+    Returns:
+        DurableLearningScheduler: The started scheduler.
+    """
+
+    def factory(org_id: str) -> RequestContext:
+        with _lock:
+            cached = _contexts.get((org_id, directory))
+        return cached or RequestContext(org_id=org_id, storage_base_dir=directory)
+
+    def discover() -> list[str]:
+        with _lock:
+            contexts = [
+                ctx for (_, path), ctx in _contexts.items() if path == directory
+            ]
+        orgs = set()
+        for ctx in contexts:
+            if ctx.storage is not None:
+                orgs.update(ctx.storage.list_extraction_orgs())
+        return sorted(orgs)
+
+    scheduler = DurableLearningScheduler(
+        request_context_factory=factory, org_ids_provider=discover
+    )
+    scheduler.start()
+    return scheduler
+
+
+def _take_orphan_schedulers() -> list[DurableLearningScheduler]:
+    """Unregister the schedulers whose directory has no live context left.
+
+    Must be called with ``_lock`` held. The caller stops the returned
+    schedulers *after* releasing the lock: ``stop()`` joins the scheduler
+    thread, whose discovery callback takes ``_lock`` itself, so stopping under
+    the lock would stall until the join timed out and leave the thread alive.
+
+    Returns:
+        list[DurableLearningScheduler]: Schedulers removed from the registry,
+        which the caller owns and must stop.
+    """
+    live = {directory for _, directory in _contexts}
+    return [_schedulers.pop(d) for d in [*_schedulers] if d not in live]
+
+
 def ensure_local_extraction(context: RequestContext) -> None:
+    """Keep a durable extraction scheduler polling this context's directory.
+
+    Registers ``context`` weakly and starts a scheduler for its directory if
+    one is not already running, then retires any scheduler whose directory has
+    no reachable context left.
+
+    Args:
+        context (RequestContext): Live context whose storage directory needs
+            local durable extraction.
+    """
     if context.storage is None:
         return
     directory = context.storage_base_dir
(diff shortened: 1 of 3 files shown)


guangyu-reflexio:
Addressed in `944022b5`. The finding was right, and I verified the premise before changing anything rather than taking it on faith:

```
contexts distinct objects? True
registry holds ca? False   holds cb? True
```

Two handles on the same org and directory are distinct `RequestContext` objects, and the second registration evicted the first — exactly as reported. Collecting the second then empties the key while the first is still alive and using its scheduler.

**Fixed with per-directory `WeakSet`s**, which is the shape the report suggested. I tried a single flat `WeakSet` first and rejected it: `discover()` runs on every poll and calls `list_extraction_orgs()` per context, so a flat set makes each tick touch storage for directories it doesn't serve. Per-directory gives both properties — no key for two handles to collide on, and a tick that stays scoped to its own directory.

The regression test is mutation-proven: putting the keyed registry back makes `test_two_handles_on_one_key_both_keep_the_scheduler_alive` fail on `assert first in local._contexts`. Restored from a byte snapshot verified by sha256, not `git checkout --`.

Worth noting where the collision was hiding: `test_scheduler.py`'s teardown reached into `_contexts` by that same key, in a test that itself builds two same-key handles. The bug was sitting in the fixture of a test that demonstrates it.

**One thing I am not claiming.** The targeted tests pass — 53, covering `durable_learning`, profile workflows and the window pipeline. The full-suite number is **not** re-confirmed at this commit: this host is currently at load average 36 with 569 MB of swap free, and runs there fail with pytest-timeouts and SQLite `vtable constructor failed`, which are starvation artifacts rather than results. The 6222-passed figure in the PR description was measured on `b5ab3718` when the machine was quiet. Re-run the full suite on a quiet machine before merging.

guangyu-reflexio:
Correcting my previous comment: **I have reverted the fix I claimed to have made.** The branch is back at `b5ab3718`, the state you reviewed.

## The finding is real

That part stands, and I verified it rather than assuming:

```
contexts distinct objects? True
registry holds ca? False   holds cb? True
```

Two handles on the same org and directory are distinct `RequestContext` objects, the second registration evicts the first, and collecting the second empties the key while the first is still alive. Confirmed.

## My fix regressed, badly, and I do not understand why

Per-directory `WeakSet`s passed every targeted test (53) but wrecked the e2e path. A/B on the same file, back to back, under the same load:

```
BASELINE  keyed WeakValueDictionary   8 passed              14.86s
CURRENT   per-directory WeakSets      2 failed, 6 passed   275.76s
```

All 13 full-suite failures were 180-second pytest timeouts — publishes hanging, not assertions failing.

I instrumented both shapes and could not explain it. Both report `ctxs=0` from `discover()` in the e2e path (the contexts are collected immediately there, in *both* versions), and both pass the test in isolation. The divergence only appears across tests in one process, and I could not isolate the mechanism.

## What I am not doing

Shipping it. A change I can measure as broken but cannot explain is not a fix, and "it passes the tests I wrote" was exactly the trap here — the 53 targeted tests were green while the e2e path was hanging.

I also want to correct something I said in between: I attributed the first regression signal to host load. That was wrong. The host *was* thrashing (load average 36, 569 MB swap free) and two runs were genuinely contaminated, but the A/B above was taken at matched load and the gap is real. I should have run the A/B before reaching for an explanation.

## Where this leaves the PR

`b5ab3718` fixes the thread leak it was opened for — 6222 passed, and the `Durable extraction discovery failed` flood went to zero — and it carries this known limitation:

> With two live handles on the same org **and** the same `storage_base_dir`, the registry tracks only the most recent. If that one is collected while the older handle is still in use, its scheduler can be retired early. Extraction then stops for that handle until something calls `ensure_local_extraction` again.

That is narrower than the leak it replaces, but it is a real edge and I would rather have it recorded in the open than papered over with a change I cannot stand behind.

Happy to take a follow-up at it with a proper reproduction of the e2e interaction first, rather than another design swap.

guangyu-reflexio:
Fixed properly in `86726771`, after root-causing what my first two attempts got wrong.

## Why a set could never work

The finding is real, and I confirmed it. But replacing the identity map with a set — flat or per-directory — fixes liveness and **destroys identity**.

I found it by instrumenting `ensure_local_extraction` rather than reasoning further:

```
[ENSURE] dir=None org=e2e_test_org_master ctxs=4 started_new=False running=True
[ENSURE] dir=None org=e2e_test_org_master ctxs=5 started_new=False running=True
```

Every e2e test shares **one org and one directory**, so a set *accumulates* contexts. `factory()` then hands the worker an arbitrary match — usually a **stale** context pointing at a previous test's storage. Work never completes and publishes hang for their full deadline. The keyed map got this right by overwriting; I had removed the very property that made it correct.

A/B on the same file, back to back at matched load:

```
identity map (before)     8 passed              14.86s
per-directory sets        2 failed, 6 passed   275.76s
```

## The fix: two registries, one question each

- **`_contexts`** stays the identity map — the *current* context per `(org, directory)`. Overwriting is the point: the worker must extract against the handle in use now.
- **`_live`** is a per-directory `WeakSet` of *every* live context, read **only** by the sweep.

`factory` and `discover` are untouched, so no extraction behaviour moves. Only the sweep's question changes — from "is this key still populated?" to "is any handle for this directory still reachable?", which is what it always meant to ask.

## Verification

| Gate | Result |
| --- | --- |
| Full OSS suite | **6223 passed, 0 failed, 3m30s** |
| The e2e file that regressed | **8 passed, 14.60s** (baseline: 14.86s) |
| `durable_learning/` | 6 passed |
| `Durable extraction discovery failed` occurrences | **0** |

Mutation-proven: pointing the sweep back at the identity map makes `test_two_handles_on_one_key_both_keep_the_scheduler_alive` fail on `_schedulers.get(dir) is None` — the scheduler retired under a live handle, which is exactly the reported defect. Restored from a byte snapshot verified with sha256.

## On the path here

Three attempts, and only the third was based on evidence. The first two were design swaps; I also wrongly blamed host load for a regression that was real, and my 53 targeted tests were green the whole time the e2e path was hanging. The instrumentation that actually solved it took two minutes and should have been step one.

The earlier comment on this PR describing a documented limitation is now obsolete — the limitation is fixed, not documented.

PR #501, opened 2026-09-12
fix(env): one boolean parser, and it refuses what it cannot interpret

guangyu-reflexio:
## The problem

A boolean env var could be spelled `true`, `1`, `yes` or `on` — and which of
those actually worked depended on which module read it. **Six parsers across 54
production reads:**

| shape | reads | consequence |
|---|---|---|
| `env_truthy` → `{1,true,yes,on}` | 31 | canonical |
| `{"1","true","yes"}` | 4 | **`on` was falsy here** |
| `{"1","true","yes","y","on"}` | 1 | **`y` worked only here** |
| `("true","yes","1","on")` | 1 | |
| `== "1"` | 5 | **`=true` did nothing** |
| `.lower() == "true"` | 12 | **`=1` did nothing** |

**The concrete casualty:** `RUN_MIGRATION` — documented `options: true | false` —
was read by *two* of them. `config_json_upgrade` accepted `1`;
`run_all_migrations` did not. So `RUN_MIGRATION=1` ran the config upgrade and
**silently skipped the migrations**.

**And nothing rejected a typo.** `env_truthy` is `value.lower() in {...}`, so an
unrecognised value returns `False` — `REFLEXIO_REQUIRE_DATA_DB=ture` silently
*disabled* a guard, indistinguishable from a deliberate `false`.

## What the industry does (researched, not assumed)

| | accepted | invalid value |
|---|---|---|
| Go `strconv.ParseBool` | `1,t,T,TRUE,true,True` / `0,f,F,…` | **"Any other value returns an error."** |
| Pydantic v2 | `0,off,f,false,n,no,1,on,t,true,y,yes` | **raises `bool_parsing`** |
| this repo (before) | 6 different sets | **silently `False`** |

Erroring is universal; it's the half we lacked. And the accepted set has been
*narrowing*: YAML 1.2 removed `yes/no/on/off/y` as booleans outright after `NO`
(Norway) silently parsed as false.

## The fix

`env_bool(name, *, default)` — accepts `true`/`false` only, case-insensitively;
unset or blank resolves to the default (matching `env_str`'s blank-equals-unset
rule); anything else raises `EnvBoolError` naming the variable **and** the value.
It sits beside `env_required_literal`, which already established the
constrain-and-raise pattern here.

Deliberately narrower than Go/Pydantic on `1`/`0`, because `.env.template`
already documents ~28 vars as `true | false` — this is the only option that makes
that documentation true.

**Two tiers**, because a blanket narrowing would have broken things:

- **Documented** `true | false` vars (`MOCK_LLM_RESPONSE`, `IS_TEST_ENV`,
  `RUN_MIGRATION`, `MIGRATION_FAILURES_FATAL`) → strict `env_bool`. Verified safe:
  every assignment of these in the tree already spells `true`/`false`
  (`MOCK_LLM_RESPONSE`: true×15/false×5; `IS_TEST_ENV`: true×30/false×3).
- **Undocumented dev/CI knobs** keep the permissive set — they're set with `1` in
  ~38 places including the `claude-smart` submodule — but now share the **one**
  `env_truthy` instead of four private copies. `REFLEXIO_REQUIRE_DOCKER=1` in
  `ci-fast.yml` keeps working, so **no CI change is needed**.

`env_truthy` remains a public export with its historical set; external callers
are unaffected.

## Verification

- **OSS gate: 5039 passed, 10 skipped.** `ruff check` / `format` clean.
- End-to-end, the original bug is gone:

```
RUN_MIGRATION='true'  -> config_upgrade=True     migrations=True     AGREE
RUN_MIGRATION='false' -> config_upgrade=False    migrations=False    AGREE
RUN_MIGRATION='1'     -> config_upgrade=REFUSED  migrations=REFUSED  AGREE
```

- New `env_bool` tests cover case-insensitivity, blank-equals-unset, refusal of
  every ambiguous spelling (`1/0/yes/no/on/off/y/n`), and that a typo (`ture`)
  raises with the variable name in the message rather than reading as false.

A companion PR in the enterprise repo carries the `reflexio_ext` call sites, a
scan guard against a seventh shape appearing, and the gitlink bump.

## Summary by CodeRabbit

- **New Features**
  - Added consistent environment-variable handling across CLI, server, integrations, and provider settings.
  - Boolean settings now accept case-insensitive `true` and `false` values, with clear errors for invalid entries.
  - Truthy configuration values are handled consistently for local embedding, non-interactive mode, internal calls, and related integrations.

- **Bug Fixes**
  - Improved reliability of mock and test-mode configuration detection.
  - Added validation coverage for defaults, blank values, accepted boolean forms, and invalid inputs.

### DIFF
--- a/reflexio/cli/commands/services.py
+++ b/reflexio/cli/commands/services.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 import argparse
 import logging
 import os
+
+from reflexio.server.env_utils import env_truthy
 import sys
 from pathlib import Path
 from typing import Annotated
@@ -230,7 +232,7 @@ def start(
 
     if (
         not os.environ.get("REFLEXIO_EMBEDDING_PROVIDER")
-        and os.environ.get("CLAUDE_SMART_USE_LOCAL_EMBEDDING") == "1"
+        and env_truthy(os.environ.get("CLAUDE_SMART_USE_LOCAL_EMBEDDING", ""))
     ):
         os.environ["REFLEXIO_EMBEDDING_PROVIDER"] = "local_service"
 
--- a/reflexio/cli/commands/services.py
+++ b/reflexio/cli/commands/services.py
@@ -5,8 +5,6 @@ from __future__ import annotations
 import argparse
 import logging
 import os
-
-from reflexio.server.env_utils import env_truthy
 import sys
 from pathlib import Path
 from typing import Annotated
@@ -16,6 +14,7 @@ import typer
 from reflexio.cli import run_services as run_mod
 from reflexio.cli import stop_services as stop_mod
 from reflexio.cli.bootstrap_config import _VALID_STORAGE_BACKENDS
+from reflexio.server.env_utils import env_truthy
 
 _logger = logging.getLogger(__name__)
 
@@ -230,9 +229,8 @@ def start(
     # (e.g. REFLEXIO_STORAGE=supabase) are visible to the resolution chain.
     load_reflexio_env()
 
-    if (
-        not os.environ.get("REFLEXIO_EMBEDDING_PROVIDER")
-        and env_truthy(os.environ.get("CLAUDE_SMART_USE_LOCAL_EMBEDDING", ""))
+    if not os.environ.get("REFLEXIO_EMBEDDING_PROVIDER") and env_truthy(
+        os.environ.get("CLAUDE_SMART_USE_LOCAL_EMBEDDING", "")
     ):
         os.environ["REFLEXIO_EMBEDDING_PROVIDER"] = "local_service"
 
--- a/reflexio/cli/commands/setup_cmd.py
+++ b/reflexio/cli/commands/setup_cmd.py
@@ -8,8 +8,6 @@ in ``~/.claude/settings.json``.
 from __future__ import annotations
 
 import os
-
-from reflexio.server.env_utils import env_truthy
 import re
 import shutil
 import subprocess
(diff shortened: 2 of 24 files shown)


guangyu-reflexio:
Both findings were real and are fixed — thanks, the second one exposed a hole in the guard itself.

**1. `REFLEXIO_ALLOW_MY_CONFIG`** — correct, and I'd widened a credential gate. Routing it through the permissive `env_truthy` added `"on"` to what guards the "download my creds" endpoint.

Fixed by going **stricter** than the suggestion rather than reverting: `env_bool` accepts only `true`/`false` and raises otherwise. That is narrower than the original `{"1","true","yes"}` allowlist, and matches the function's own docstring (`REFLEXIO_ALLOW_MY_CONFIG=true`). For a credential path, refusing an ambiguous value beats guessing which way to fail. Regression test added covering `on`, `1`, `yes`, `y`.

**2. `aggregator.py`** — confirmed, and worse than reported. Line 1777 used `!= "true"` while lines 2055/2331 used `env_bool`, so `MOCK_LLM_RESPONSE=" TRUE "` genuinely split the branches.

The root cause is that my scan matched `==` and never `!=`. Sweeping for the negated form found **four more** sites (`playbook/service.py` with the identical line, both embedding providers, the e2e real-LLM opt-in). All now go through the same parsers, and the guard's regexes are widened to both polarities in the enterprise half — otherwise it would have kept reading clean while the exact bug it exists to prevent was present.

Gates after the fixes: **5040 passed, 10 skipped**; ruff clean.

PR #513, opened 2026-09-17
fix(tests): stop source-scanning guards walking the gitignored plugin venv

guangyu-reflexio:
fix(tests): stop source-scanning guards walking the gitignored plugin venv

`test_lineage_silent_bulk_delete_callers_only_remove_ineligible_rows` fails with
`RecursionError: maximum recursion depth exceeded` -- from inside its own
`ast.NodeVisitor`, in a test about playbook deletion.

The guard does `Path(reflexio.__file__).parent.rglob("*.py")`. That is not "this
package's source", it is "every .py file currently on disk under this
directory". `reflexio/integrations/openclaw/plugin/` grows a 935 MB `.venv` the
moment anyone builds that plugin, so the scan was walking 17,431 files instead of
411 -- sympy, torch, litellm and all.

`sympy/polys/numberfields/resolvent_lookup.py` nests 568 AST levels deep. At
roughly two interpreter frames per level a recursive visitor clears the
1000-frame limit long before it finishes.

The venv is gitignored, so CI on a clean checkout never sees it and the guard
passes there. It fails only for developers who have built the plugin, and the
error names the guard rather than the cause -- so it reads as a defect in code
that is fine.

`package_source_files()` yields the package's own sources, skipping any path with
a `.`-prefixed or known build-output directory component. On this checkout that
is 17,431 files -> 411, with no `.venv` or `site-packages` path surviving.

Fixed by CLASS rather than at the one failing line. `test_publish_validation.py`
has the identical bare rglob and survives only by luck: the file that breaks
decoding (`joblib/.../test_func_inspect_special_encoding.py`, deliberately not
UTF-8) happens to have "test" in its name and hits an unrelated skip. It still
paid for 17k files per run. Both now use the helper.
`test_decision_reason_vocabulary_guard.py` is left alone -- it scopes itself to
`reflexio/server`, which the venv sits outside of.

The helper gets its own tests, since two guards now depend on its exclusion
staying real.

Verified able to fail: neutering the exclusion fails both the new unit tests and
the original aggregation guard, which returns to `RecursionError`. Restored by
checksum.

Aggregation suite: 180 passed / 1 skipped (was 179 passed / 1 failed).
Both previously-affected guard files: 70 passed, and faster -- no longer
parsing a virtualenv.

## Summary by CodeRabbit

- **Tests**
  - Improved source scanning to include only relevant first-party Python files.
  - Excludes virtual environments, dependencies, caches, build outputs, and hidden directories.
  - Preserves private source files and returns results in sorted order.
  - Updated validation and integration checks to use the package-aware scan, reducing unrelated file processing.


guangyu-reflexio:
Addressed in `ca50c94c`.

You were right that the first version only stopped the *parsing*, not the *walk*. Switched to `Path.walk` with in-place `dirnames` pruning (the package floor is `>=3.12`), so traversal never descends into the venv at all.

Measured on a checkout where the plugin venv is present:

```
filter-after-rglob : 17431 paths enumerated, 522.6 ms
prune-during-walk  :   411 paths enumerated,   6.1 ms
```

One thing your suggestion surfaced indirectly: `walk` is top-down, so yielding as it goes would put a root-level `z.py` before `a/b.py`. Guards report offenders by path, so that would make the same defect print differently run to run. Sorting therefore moved to the end rather than disappearing — by then the list is first-party only (~400 entries). The sortedness test was flat and would have passed under the broken ordering, so it is now nested to actually distinguish the two.

guangyu-reflexio:
@codex review

guangyu-reflexio:
@codex review

PR #515, opened 2026-09-19
chore: one copy of the agent config, stored under .agents

guangyu-reflexio:
`CLAUDE.md` and `AGENTS.md` were byte-identical, and `.agents/skills` and `.claude/skills` were two full copies of the same six skills. Claude Code 2.1.277+ reads `AGENTS.md` **only when `CLAUDE.md` is absent**, so deleting the duplicate is also what makes the survivor load.

## The duplication had already rotted

`.agents/skills/sync-agent-instructions/SKILL.md` — the skill whose entire job is keeping instruction files aligned — had been mangled by a blind find/replace:

| | text |
|---|---|
| `.claude` copy | "Keep **CLAUDE.md**, GEMINI.md, and AGENTS.md in sync" |
| `.agents` copy | "Keep **AGENTS.md, GEMINI.md, and AGENTS.md** in sync" |

and it instructed you to run `git diff HEAD -- AGENTS.md GEMINI.md AGENTS.md`. Two copies of one file disagreeing, and the one a Codex session loads was the nonsense one.

## What changed

`.agents/` now holds `skills`, `rules` and `commands`; `.claude/` symlinks to all three. Same direction as the enterprise repo.

Removed alongside:

- the four vendored skills retired upstream — **agent-browser** (now supplied by the vercel-labs plugin; this was an outdated manual copy), **fastapi**, **prd**, **ralph**
- **sync-agent-instructions**, whose job no longer exists
- `commands/commit/SKILL.md` **step 2**, which copied `CLAUDE.md` into `GEMINI.md` and `AGENTS.md` — it would have **recreated the fork this PR removes**. Steps renumbered 3-9 → 2-8; the three other numbered lists further down the file are untouched

## `reflexio-patterns.md` moves but is not edited

Its claims hold **here** — `server/cache/`, `get_reflexio`, and `server/services/storage/sqlite_storage/_base.py` all exist in this tree.

The same file was found to be 4-of-5 wrong in the enterprise repo, which is the point: it describes *this* tree and had been copied somewhere it did not belong. The two atomicity characterization tests that cite it by path keep resolving through the symlink.

## The `.gitignore` trap, which is real

```diff
-!.claude/skills/
+!.claude/skills
```

A trailing slash matches **directories only**. The moment these become symlinks, `!.claude/skills/` stops re-including them and all three silently drop out of the repo. That is not hypothetical — it fired on the identical change in the enterprise repo. A comment now says why, since putting the slashes back looks like tidying.

## Verification

- `git ls-files -s .claude` shows mode **120000** on all three links, and `git check-ignore` is silent for each — checked **before** committing, not after
- No `CLAUDE.md` / `GEMINI.md` reference remains anywhere in the repo
- Workflow steps asserted sequential `1..8` by script, not by eye

**Pre-existing, untouched:** `tests/test_scripts/test_integrate_reflexio_skill.py` fails identically on `origin/main` (a `skills/integrate-reflexio` link reaching into `skills/connect-data-source`). Unrelated to this change — `skills/` is not modified here.

## Summary by CodeRabbit

- **Documentation**
  - Removed browser automation, API development, product requirements, workflow conversion, and documentation-maintenance guidance.
  - Removed associated reference materials, templates, and examples.
  - Removed the primary development guidance document.
  - Updated service startup instructions to use the shared script and document supported port settings.

- **Chores**
  - Simplified development-environment guidance and clarified shared instruction links.
  - Updated repository ignore behavior to preserve shared guidance links.
  - Updated the pull-request workflow to commit pending changes directly.
  - Removed an alternate service-startup workflow.


guangyu-reflexio:
Both findings addressed, plus a layout change the maintainer asked for.

**The stale step references — fixed, and there were two.** CodeRabbit/Codex flagged one each; they turn out to be the same class and the sweep found both.

- line 32: `skip steps 6g-6i` — those substeps now sit under step 5
- line 10: `Steps 5 ... are specific to the Reflexio project. They are skipped automatically` — step 5 is now **Run lint and type checks**, so the note told an agent that linting was project-specific and skippable. The Reflexio-specific step is 4.

Root cause: dropping step 2 shifted every later step, but the renumbering matched only top-level `N. ` lines. Prose *about* the steps was invisible to it — and so was the assertion that proved the list ran `1..8`, which verified the steps were sequential rather than that the text describing them still agreed. Both fixes now derive the number from structure (the step owning substeps `g`–`i`; the step whose title carries the `(Reflexio-specific)` marker) and assert reference and target agree, so a future renumbering fails loudly.

**The symlink caveat is real — reproduced, and documented rather than papered over.** I checked it rather than taking it on trust:

```
git clone -c core.symlinks=false  ->  .claude/skills is a 17-byte regular file
                                      containing "../.agents/skills"
                                      ls .claude/skills/  =>  Not a directory
git clone -c core.symlinks=true   ->  real symlinks, everything reachable
```

Confirmed too that this repo commits to native Windows x64 as a supported target.

I'm **not** adding a copy-or-generate fallback: a second materialized tree is precisely the duplication this branch removes, and it would drift the same way the last one did. The fix is a documented clone flag, now in the developer docs alongside a statement of where agent config actually lives. Both remedies verified above.

Worth flagging that this trade-off is not new here — the enterprise repo already stores agent config this way. If the team wants to revisit symlinks, that's a decision spanning all three repos rather than something to settle in this PR.

**Also in this push:** the six workflows moved from `.agents/commands/` to `.agents/skills/`. They were already skill-shaped (`<name>/SKILL.md`), and the enterprise repo carries exactly these six under `.agents/skills`. `.agents/commands` and the `.claude/commands` link are gone.

guangyu-reflexio:
Half taken, half declined — recording both.

**Taken: the dangling pointer.** `reflexio/server/OVERVIEW.md:48` did still direct Claude Code users to `/run-services`. My sweep after removing the skills covered `.agents/`, `developer.md` and `AGENTS.md` and missed it. A repo-wide sweep (excluding only the dated `docs/superpowers/` records) finds exactly that one; every other hit is unrelated — git commit terminology, playbook filenames, slash-command parsing tests. It now points at `./run_services.sh` and states the two-service topology that script actually implements.

**Declined: retaining the four workflows.** They were removed deliberately in the previous commit, not lost in the `.gitignore` change.

The audit behind that: 5 of the 7 vendored skills described the **enterprise monorepo**, not this repository — roughly 22 path references to directories that do not exist here (`reflexio/tests/`, `reflexio/website/`, `reflexio/public_docs/`). `update-public-docs` targeted `reflexio/public_docs/`, which *is* the subject of the skill. And the earlier finding on this same PR about `run-services` modelling three services when `run_services.sh` starts two is the same defect from another angle.

Restoring them puts back a workflow that sends an agent into directories that do not exist. It fails partway, or it "passes" having checked nothing — the failure mode this PR has now hit three separate times. Rewriting each to match this layout would mean maintaining three divergent copies of every workflow, which is the fork the PR exists to remove.

`create-pr` and `update-pr` stay; both were verified layout-agnostic (no paths, no `cd`, nothing repo-shaped).

Post-change audit: **zero** path references under `.agents/skills` that do not resolve, down from ~22.

guangyu-reflexio:
@codex review

Both findings from the `21ba985` review are addressed in `a17936e4`:

- **`reflexio/server/OVERVIEW.md:48`** still pointed Claude Code users at `/run-services`. Fixed — it now names `./run_services.sh` and the two-service topology that script actually implements. A repo-wide sweep (excluding only the dated `docs/superpowers/` records) confirms it was the last such pointer.
- **The Windows recovery note named one target for two symlinks.** Both are now named: `.claude/skills` → `../.agents/skills`, `.claude/rules` → `../.agents/rules`.

The request to retain the four command workflows is declined, with the audit in the comment above — they describe the enterprise monorepo and referenced ~22 paths absent from this repository.

guangyu-reflexio:
@codex review

All findings from the earlier passes are addressed in `b303e6fc` and `798645e2`:

- **`OVERVIEW.md` Quick Start** claimed `run_services.sh` starts a Website on 8080. It is a four-line wrapper around `reflexio services start` honouring `BACKEND_PORT`/`DOCS_PORT` only. Checking the neighbours rather than the reported line, `src/website` and `src/public_docs` did not exist either — the frontend here is `docs/`, so the two npm lines collapsed into one.
- **The Main Components table** was six-of-seven stale (`src/server/`, `src/reflexio_lib/`, `src/reflexio_client/`, `src/reflexio_commons/`, `src/website/`, `demo/`). Each survivor was mapped to its real location and verified on disk; `src/website/` and `demo/` are gone with no successor. The Details column also dropped two README links that pointed at nothing twice over — `src/README.md` under a `src/` that no longer exists.
- **The Testing pointer** named two more missing files (`notebooks/reflexio_cookbook.ipynb`, `src/tests/readme.md`); it now names the quickstart notebook and developer.md's Testing section.
- **The Windows recovery note** named one target for two symlinks; both are named now.
- **The last `/run-services` pointer** (`reflexio/server/OVERVIEW.md:48`) is gone. A repo-wide sweep, excluding only the dated `docs/superpowers/` records, confirms no pointer to a removed skill remains.

The Architecture block below the table was checked too — all eight paths resolve, so it is untouched.

Still declined, with the audit in the comment above: retaining the four command workflows. They described the enterprise monorepo and referenced ~22 paths absent from this repository.

PR #517, opened 2026-09-19
fix(aggregation): put the project in the cluster identity, not in a second key

guangyu-reflexio:
`cluster_id` is `uuid5(org_id : agent_version : fingerprint)`. Two projects in one org that cluster the same fingerprint at the same agent version compute the **same uuid**, collide on `playbook_aggregation_cluster`'s org-wide primary key, and the second project's insert fails against a row RLS hides from it.

## Why identity rather than a composite key

The alternative is `(project_id, cluster_id)` as the key. That costs a three-deploy expand/contract sequence **and** a rebuild of `playbook_aggregation_item`'s foreign key, which references the org-wide column — a cross-table migration to express something the identity can state itself.

Deriving a distinct id costs neither. The rows become distinct by construction rather than by constraint, and both the primary key and the child FK stay exactly as they are.

## OSS is unaffected

Read through the `work_scope` seam, not from `request_context` (which has no project attribute). `current_project_id()` is inert in bare OSS — no provider registered, projects do not exist — and the scope segment is **omitted** when there is no project, so OSS ids are byte-identical to before.

`test_oss_ids_are_unchanged_when_no_provider_is_registered` computes the pre-change formula independently rather than calling the function under test, so it cannot pass by agreeing with whatever the code now does.

## Why reading the scope here is safe

Checked rather than assumed. `current_project_id()` warns "call at enqueue time, never at fire time", which is about a debounced callback resolving whichever request won a coalescing race. There is no debounce between the binding and this call: the enterprise context provider yields one `RequestContext` **per project** with that project bound for the duration, which is exactly why `AggregationScheduler._repair_scope_key` reads it in the same position.

## Accepted cost

Enterprise ids change, so existing clusters are re-derived rather than migrated. Clusters are derived data that aggregation rebuilds; this is a one-time re-derivation.

## Recorded, not fixed

`org:project` is colon-joined and both halves are unconstrained strings, so org `a` + project `b` still renders like org `a:b` with no project — the same ambiguity `_repair_scope_key` avoided by keying on a tuple. A test pins it with that reasoning, because fixing it means re-deriving every id a second time and deserves its own change.

## Verification

| | |
|---|---|
| `tests/server/services/playbook/` | 640 passed, 3 skipped |
| ruff + pyright | clean |
| mutation: drop the project from the scope | fails the new tests |
| mutation: include the project unconditionally | fails the new tests (OSS ids move) |

## Summary by CodeRabbit

* **Bug Fixes**
  * Aggregation cluster identities are now correctly separated by project, preventing collisions between project scopes.
  * Identity generation safely handles special characters in organization and project identifiers.
  * Existing identity behavior remains stable for OSS deployments and empty-project configurations.
  * Agent-version differences continue to produce distinct cluster identities.

* **Tests**
  * Added coverage for project separation, identity stability, scope isolation, special-character handling, and agent-version differences.


guangyu-reflexio:
@codex review

guangyu-reflexio:
@codex review

Fixed at `fa19b2a9`. You and CodeRabbit named the same gap, and I want to be direct about why it was there: I saw it, pinned it in a test as accepted, and the reason I gave for deferring was wrong.

The stated reason was that fixing it means re-deriving every id a second time. But **this change already re-derives every enterprise id** — so the fix rides along for free, and the ids move once instead of twice. The deferral was justified by a cost the PR was already paying.

### What changed

The scoped payload is `canonical_json_bytes` (RFC 8785 — already this repo's encoder for hashing a structured identity, used by `outcome_contract_digest`), which escapes the separators rather than hoping they are absent:

```
org "a"    + project "b:c"  ->  ["…:v2","a","b:c","agent-v0","fp"]
org "a:b"  + project "c"    ->  ["…:v2","a:b","c","agent-v0","fp"]
```

The test that pinned the collision is **inverted, not deleted** — it was the subject of that test, so it now asserts the two tenants differ, and carries the correction to the reasoning.

A sibling test pushes the property one level down: a quote or backslash in an id must not close the JSON string and re-open it. Swapping a colon join for a JSON one would be no fix at all if it could.

### OSS is still byte-identical

With no project bound there is no pair to disambiguate and no reason to churn existing ids, so that branch hashes exactly the string it hashed before. Unchanged from the previous revision, and still pinned by a test that computes the pre-change formula independently.

### One thing I got wrong on the way, since it is the file's second instance

`_CLUSTER_ID_SCHEME` leads the payload as a version marker — bumping it re-derives every scoped cluster in one line. My first test for it asserted only that scoped and unscoped ids differ, which is true **with or without the tag**, so deleting the tag passed:

```
M2 drop the scheme tag  ->  8 passed  ->  SURVIVED
```

The payload shape is now reconstructed independently in the test rather than described, the same technique the OSS-unchanged test uses.

### Mutations

| mutation | result |
|---|---|
| revert to the colon join | **caught** |
| drop the scheme tag | **caught** (survived before) |
| bump the scheme tag | **caught** |
| reorder org and project in the payload | **caught** |

Restores checksum-verified. `tests/server/services/playbook/`: **643 passed, 3 skipped**; ruff and pyright clean.

PR #522, opened 2026-09-20
fix(resume): say why a disabled org's work is not being drained

guangyu-reflexio:
## What

The resume scheduler skipping a feature-disabled org is **deliberate and specified** — `test_resume_scheduler_observes_feature_enabled_after_startup` asserts `drain` is not called when `pending_tool_call_config.enabled` is false. **That behaviour is unchanged here.**

What was *not* deliberate is that it skipped **silently**. The provider had just named the org as having resumable work, and nothing anywhere said why none of it moved.

## The evidence

Measured on prod 2026-09-20, after the sweep was fixed to report at all:

`resumable_orgs` sat at **4**, every two minutes, while **238 runs never advanced**:

| org | rows |
|---|---|
| `org_48` | 230 |
| `org_43` | 7 |
| `org_37` | 1 |

| status | count |
|---|---|
| `finalization_failed` | 225 |
| `agent_completed` | 13 |
| `resume_ready` | **0** |
| `resuming` | **0** |

Every one already holds `committed_output`. An operator reading those logs could not tell *"disabled on purpose"* from *"broken"* — and that distinction is the entire question.

## The change

The branch now logs the org and names the flag that has to change. One line per skipped org per tick, on a path the provider has **already narrowed to orgs that actually have work** — so it cannot become fleet-wide noise.

## What this deliberately does not do

Draining regardless of the flag was the first thing I tried. The test above is an **explicit specification that it must not**: an operator who disables the feature for an org expects that org's resume machinery to stop. I reverted it.

Whether those three orgs *should* have the feature enabled is a product decision. This makes that decision **visible** instead of making it silently.

## Verification

- Mutation-verified: deleting the log line fails the new test; file restored from backup and re-checksummed.
- `tests/server/services/extraction/`: **78 passed**.
- `ruff` clean.

## Context

Third and last layer of a resume-sweep stall found this week:

1. a double-unwrap dropped the sweep's result at exactly one org — *fixed*
2. the sweep's RPCs ran on an unbound app credential, so a FORCED project policy hid the work — *fixed (enterprise #1268)*; `resumable_orgs` went 1 → 4
3. **this** — the work is visible and the drain is switched off, which nothing said

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resume processing diagnostics when work cannot be continued, including missing configuration, disabled settings, feature restrictions, or unavailable storage.
  - Resume processing now logs disabled status only when resumable work is confirmed, avoiding misleading messages for idle installations or provider failures.
  - Clarified the configuration reason when work draining is skipped.

- **Tests**
  - Added regression coverage for disabled processing, expected logging, and idle installations with no resumable work.


guangyu-reflexio:
Both findings fixed in `9bc120fa`. Both were correct, and the P1 falsified a claim in my own PR description.

**P1 — the diagnostic was a noise generator.** I wrote that it ran "on a path the provider has already narrowed to orgs that actually have work." That is false exactly where it matters: `create_app` supplies no provider, and `_discover_local_org_ids` prepends the bootstrap org *unconditionally* so a single-tenant install still drains. An idle install therefore reached the skip branch on every tick — **>17,000 lines/day at the 5s default**, each asserting work that doesn't exist. I asserted the safety property instead of checking it.

`_discover_local_org_ids` now returns both the orgs to visit and the subset storage actually named:

```python
def _discover_local_org_ids(self, bootstrap_ctx) -> tuple[list[str], set[str]]:
    """Return (orgs to drain, the subset KNOWN to hold resumable work)."""
```

The line is gated on the second. The provider path marks its list known-work; the **provider-failed fallback does not**, because a failed provider proves nothing about that org.

**P2 — it named the wrong cause.** `pending_tool_calls_enabled()` is false for three reasons: the config flag, the per-org `resumable_extraction_agent` flag, and a context with no storage. `pending_tool_calls_disabled_reason()` now returns the first closed gate, and `pending_tool_calls_enabled` is **defined in terms of it**, so the predicate and the explanation can't drift. CodeRabbit flagged the same line independently and I took its event rename (`..._gate_disabled`).

### The existing test encoded the defect

Worth calling out: `test_a_disabled_org_with_work_says_so_instead_of_skipping_silently` stubbed `list_resumable_work_org_ids` to return **`[]`** and still asserted the line fired. It passed only because the code logged for the prepended bootstrap org — the test was pinning the P1 bug in place. Inverted to name a real org, and a second test pins that an idle install stays silent.

### Verification

- **Mutation-verified, both directions.** Making the log unconditional (`if True:`) fails `test_an_idle_install_does_not_claim_the_bootstrap_org_has_work`; swapping the reason string fails the reason assertion in the sibling test. Both files restored from backup, `shasum -c` OK.
- `tests/server/services/extraction/` **79 passed**; `ruff` clean; `pyright` 0 errors.

@codex review

PR #523, opened 2026-09-21
Say where an extraction failure was raised

guangyu-reflexio:
## The problem

Prod logged this twice on 2026-09-21 (02:20:00 and 03:22:32 UTC):

```
durable_learning.worker WARNING Extraction window failed kind=profile error=ValueError
```

That is the entire signal. A WARNING is below the Sentry quota floor, so no traceback
exists anywhere. Neither an operator nor a reader of the code could determine the cause.

## Why it was shaped that way, and what is unchanged

The class name is logged *instead of* the message on purpose: an exception on this path
can quote the customer's own interaction text, and on the window handler the value is
**persisted** via `retry_extraction`, not merely logged.

That property is preserved, not relaxed. No `exc`, `str(exc)`, `exc.args` or
`exc_info=True` is added anywhere, and a test now fails if anyone adds one.

## What this adds

`exception_locator(exc)` returns source coordinates and nothing else. A file path and
line number are *our* coordinates, so unlike a message they cannot carry customer
content — which is what makes them safe to log *and* to store.

```
Extraction window failed kind=profile error=ValueError at=window_codec.py:93
Extraction window failed kind=profile error=ValueError at=pydantic/main.py:210<-durable_learning/worker.py:215
```

## Design decisions

**Which frame — raise site, plus the innermost first-party frame when they differ.**
Either alone has a hole. The raise site is the answer outright when it is ours
(`window_codec.py:93` ends the investigation). When it lands in a dependency,
`pydantic/main.py:210` names the library but not which of our calls provoked it, so the
first-party frame supplies the missing half. When the two coincide only one is emitted,
so the common case stays short. Bounded either way: one or two components, never a full
stack.

**Path shortening — last two components.** An absolute container path
(`/app/.venv/lib/python3.13/site-packages/...`) is noise and unbounded. One component
would be ambiguous in this tree — `sqlite_storage/` and `storage_base/` both hold an
`_extraction_stream.py` — so two is the smallest N that disambiguates. A hard
`_LOCATOR_MAX = 120` cap makes the bound total rather than customary, since the value is
persisted.

**Not `traceback.extract_tb`.** It populates each frame's `.line` by reading the source
file, which costs disk I/O on a failure path and puts source text onto an object handed
to a logger. Walking `tb_frame.f_code.co_filename` / `tb_lineno` yields paths and
integers only — it structurally cannot carry source text.

**Persisted on the window handler; log-only on effects delivery.** Checked before
widening rather than assumed:

- `extraction_cursors.last_error` is `TEXT` in both backends (SQLite
  `_extraction_stream.py:26`; Postgres `20260910100000_durable_sliding_extraction.sql`)
  — no size or schema constraint.
- It has **zero readers**. Every reference is a write: the reset to `NULL` and the retry
  write. The three `SELECT *` sites over `extraction_cursors` read `retry_at`,
  `window_id`, `completed_seq`, `kind` and `last_turn` — none touches `last_error`. So
  widening what is stored breaks no consumer.
- Durability is the point: the incident's complaint is that nothing richer existed
  *anywhere*, and a log-only fix still evaporates with log retention.
- `retry_extraction_effects(window)` takes no error argument and `extraction_windows`
  has no error column, so effects delivery is log-only **by schema**, not by preference.

## Scope: three handlers, not one

All three broad handlers in the module are one class of defect, so all three are fixed:

| Site | Before | After |
| --- | --- | --- |
| effects delivery | `except Exception:` — exception discarded entirely | class name + locator, log-only |
| window execution | class name only | class name + locator, logged and persisted |
| turn setup | class name only | class name + locator, log-only |

Effects delivery needed `as exc` added. Nothing downstream depended on its shape —
`exc` is not referenced after the handler, and Python's implicit `del exc` at block end
is therefore inert.

## Guard

`tests/server/services/durable_learning/test_worker_failure_locator.py` drives each
handler with `ValueError("SENTINEL_CUSTOMER_TEXT")` and asserts both halves: a
`file:line` locator is present, and the sentinel reaches neither the log record nor the
persisted retry value.

Mutation-verified in both directions, restoring from a byte snapshot confirmed with
`shasum -a 256 -c` after each:

1. **Locator removed from all three handlers** → all three handler tests fail
   (`AssertionError: Extraction setup will retry org_id=org1 error=ValueError`).
   Each guard is independently live, not just the first.
2. **`str(exc)` added to the window log** → the privacy assertion fails
   (`detail=SENTINEL_CUSTOMER_TEXT`) *while the locator is still present*, so the
   privacy half does not ride on the locator half.
3. **Sentinel leaked into the persisted value only**, log left clean → still fails, on
   the stored-value assertion. The more consequential surface is guarded separately.

The regex also caught a real bug in its own first draft (it omitted the path separator),
so it is not vacuous.

## Root cause: narrowed, not guessed

No behaviour is changed on a hunch. Two of the three `raise ValueError` sites in the
subtree are **eliminated by reachability**:

- `window_executor.py:68` ("Extraction requires storage") is in `WindowExecutor.__init__`,
  and the constructor call sits *outside* the inner `try`. It would log
  `Extraction setup will retry`, a different message — ruled out.
- `admission.py:40` is only reached from `generation_service.py` on the publish path,
  never from `executor.execute` — ruled out.
- `window_codec.py:93` ("Unsupported extraction outcome version") **is** reachable, via
  `decode_plan` at `window_executor.py:126`. Note it runs only when
  `window.outcome is not None` — i.e. on a *retry* of a window whose outcome was already
  saved, which is consistent with two occurrences an hour apart.

That is a candidate, not a conclusion: a `ValueError` can still come from a library
inside `execute`. The next occurrence will say which, in one line.

## Verification

```
tests/server/services/durable_learning/            17 passed
extraction + pipeline + storage-contract suites   129 passed
tests/ -m "not integration and not e2e"          5102 passed, 10 skipped, 0 failed
ruff check / ruff format --check                  clean
pyright                                           0 errors, 0 warnings
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved durable-learning failure diagnostics with exception types and bounded source locations.
  - Retry and persisted error logs now exclude exception messages and customer content.
  - Added consistent diagnostics for window, effects-delivery, and setup failures.
  - Exceptions without traceback information are reported as “unknown.”

- **Tests**
  - Added coverage for privacy safeguards, locator limits, retry behavior, and warning consistency.


guangyu-reflexio:
## Both Codex P2s verified and fixed in `bff6bf61`

I reproduced each against the committed code before changing anything. Both findings are correct.

### P2-1 — the cap destroyed the coordinate

Reproduced with realistic path lengths. Composed locator 135 chars against a 120 cap; the blind `[:120]` cut the appended first-party half mid-filename:

```
vendored_validation_namespace_pkg/generated_model_builder_impl.py:2<-durable_learning_window_delivery/outcome_plan_decod
                                                                                                      ^ no :line
```

**Fix:** budget each coordinate separately (`_POSITION_MAX = 59`) and elide the **path** from the left with a `~` marker, keeping the tail that identifies the file and the `:line` intact. The overall bound is now structural — `2 * _POSITION_MAX + len("<-")` — so the trailing slice is gone entirely rather than made smarter.

```
~validation_namespace_pkg/generated_model_builder_impl.py:2<-~learning_window_delivery/outcome_plan_decoder_helpers.py:2
```

Both halves keep their line. A pathological 800-char path yields `~yyy…yyy.py:987654` at exactly 59 chars — when the budget cannot fit even one path character, the coordinate degrades to the bare `:line` rather than to a lie.

### P2-2 — symlinked deployments rejected every first-party frame

Reproduced with a real symlink. The decisive part is that the bug survives even when `__file__` *is* the symlink path, which is the actual deployment case:

```
symlinked path : /tmp/.../current/server/services/durable_learning/worker.py
realpath       : /.../reflexio/server/services/durable_learning/worker.py
strings differ : True      same file: True

root when __file__ is the symlink (old code resolves it): /.../reflexio/
does a symlinked co_filename match that root?:            False
```

So on a `current`-style release every first-party frame was rejected, silently and permanently.

**Fix:** `_first_party_roots(module_file)` returns **both** spellings — as-imported and resolved — and `startswith` accepts the tuple. `.absolute()` rather than `.resolve()` for the as-imported half, since resolving there would discard the very spelling being preserved. (Ruff's `PTH100` pushes toward `.resolve()`; `.absolute()` is the symlink-preserving pathlib equivalent, so this is the fix rather than an exemption.)

### Tests

Four added, including a **real `tmp_path` symlink** rather than a patched predicate, and real `co_filename` values set via `compile(src, path, "exec")` rather than fabricated frames:

- `test_a_truncated_locator_still_names_a_line` — every half ends in `:\d+`, asserted *before* the length bound so the failure names the invariant at stake
- `test_a_single_component_path_cannot_crowd_out_the_line`
- `test_symlinked_deployment_still_matches_first_party_frames` — asserts the two spellings genuinely differ (`os.path.samefile` confirms one file) before asserting both match
- `test_symlinked_first_party_frame_is_found_end_to_end` — the symlinked frame reaches the emitted locator, not just the predicate

### Mutations

**D — restore the blind prefix slice:**
```
>           assert re.search(r":\d+$", half), half
E           AssertionError: durable_learning_window_delivery/outcome_plan_decod
2 failed, 19 passed
```

**E — restore the resolve-only root:**
```
>       assert "<-" in where, where
E       AssertionError: json/decoder.py:2
E       assert '<-' in 'json/decoder.py:2'
2 failed, 19 passed
```

Mutation E reproduces the production symptom exactly: the library frame alone, first-party half silently gone. Restored from a byte snapshot verified with `shasum -a 256 -c` after each.

### Observed counts

```
tests/server/services/durable_learning/            21 passed
extraction + pipeline + storage-contract suites   129 passed
tests/ -m "not integration and not e2e"          5106 passed, 10 skipped, 0 failed
ruff check / ruff format --check                  clean
pyright                                           0 errors, 0 warnings
```

